PR #18, opened 2015-01-09
Fix short-circuit logic in fuse_fiss_utils

ryankelly-uiuc:
In fuse_2_patches, the code evaluates:

`writing_long .and. all(csite%dmean_can_prss > 10.0)`

`csite%dmean_can_prss` will not be allocated unless writing_long is true,
so if writing_long is false and the second expression is evaluated, a segfault
results.

Apparently, different compilers handle these compound logic statements
in different ways. I assume that for many folks the logic is short-circuited
so that whenever writing_long is false the all(...) expression is never
evaluated, because otherwise this probably would have come up before.

I always get the segfault, however. Changing the code to explicitly check
the first expression before moving on to the second fixed it for me and shouldn't
affect anyone who had it working already.

(There are two other analogous `if()` statements that were changed the same way.)

### DIFF
--- a/ED/src/utils/fuse_fiss_utils.f90
+++ b/ED/src/utils/fuse_fiss_utils.f90
@@ -4679,7 +4679,8 @@ module fuse_fiss_utils
       !------------------------------------------------------------------------------------! 
       !    Daily means.                                                                    !
       !------------------------------------------------------------------------------------! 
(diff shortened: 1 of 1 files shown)


rgknox:
Ryan, awesome commit.  Thanks for submitting. Will pull.

PR #33, opened 2015-03-13
Option to select CBR scheme

crollinson:
Major Changes:
1) Carbon Balance is based on running sums of actual carbon balance and maximum potential carbon balance (structural_growth.f90, line 320-325)
2) Implement CBR_SCHEME (new ED2IN option) where the user must now specify how they want CBR (carbon balance ratio) to be calculated.  Currently we have implemented 3 options:
                    0 -- Single stress.  CBR = cb/cb_mlmax 				               
			             cb_mlmax is the carbon balance in full sun and no moisture    
			             limitation							                           
                    1 -- Co-limitation from light and moisture  			           
			             CBR_LIGHT = cb/cb_lightmax and CBR_MOIST = cb/cb_moistmax	   
			             CBR_LIGHT and CBR_MOIST are then weighted according to 	   
			             DDMORT_CONST (below)					                       
                    2 -- Leibig’s Law of the Minimum — Limitation from either light or 
			             moisture depending on which is lower at a given point in time 
			             CBR = cb/max(cb_lightmax, cb_moistmax)			               

---------------------------------
minor changes:
-- fixed mainline typo with cdrag in canopy_struct_dynamics


DanielNScott:
Hmmm, it looks like there are a few whole files getting their contents deleted, is that intentional? I mean the ED2IN being deleted probably isn't, but i don't know about the others?

crollinson:
@DanielNScott The ED2IN should have changes to it -- CBR_SCHEME now needs to be specified in it -- and if it was deleted, there should be a new one to replace it.  I think there's some changes in capitalization maybe that makes it think it's a whole new file?  I didn't intentionally delete any files, although I think I saw something about a soils file somewhere getting deleted (run/soil_depths.dat), but I think that happened when I either pulled from the mainline or from @ryankelly-uiuc.

I do know that the branch I submitted the pull request for compiles and runs on the BU server.

rgknox:
Once the ED2IN thing is sorted out, I will get this thing on the test suit
and compare to the changes prior to the SMP stuff.
Christy, I also noticed some edits related to snowfac and cdrag.  Can you
confirm that these are targeted modifications (ie with your intent).

I noticed a change to snow_fac that was overwritten with the original
method a few lines later.
You also added an initialization to cdrag prior to its calculation in the
icanturb 2 and 3 specific calculation.  It looks like it will just get
overwritten anyway (as you have to be in icanturb type 2 or 3 at this
point).  It looks like we don't need that line, but if we do, the logic
should hold for the 8byte clone of the code.

ryankelly-uiuc:
I think I'm the culprit for missing files in the /run directory... I reorganized mine for personal use and meant for the changes to be gitignored but looks like they made it into Christy's code. I can tell because my run scripts are in there too. Actually, I didn't even remember that there's supposed to be an ED2IN in .../run/. The one we modified to include the new CBR_SCHEME flag is in .../Templates/Templates.

@crollinson , if you restore the run directory to match the mainline and just copy over our new ED2IN changes that should clear things up a lot. If you want me to do this I can, but I don't know if I can commit changes to a pull request you started. 

I'm not sure about the include.mk files, but my guess is they should be removed from this pull request and left unchanged. And I definitely don't know nuthin about no snow_fac edits.

crollinson:
@rgknox 
I did mean to set the snowfac back to the original usage since I discovered it was not the source of an error I had been receiving.  The initialization to cdrag was fixing a typo/bug in what someone (I think maybe you?) did when I pulled the mainline version prior to submitting the pull request.  Without the initialization, the code wouldn't compile.  I thought about deleting it, but didn't want to delete things incase there was a reason the way things were.

I think I've now replaced the run files, with what was there before with the exception of the ED2IN, which is the new one with CBR_SCHEME specified.

mdietze:
@crollinson could you clarify on the snowfac -- I understand that this was not the ultimate source of your error, but are you now concluding that the original code was in fact correct? It seemed to me that the original version still convolved the ideas of snow depth relative to canopy depth and snow fractional cover, such that your revised version was more correct and thus shouldn't be reverted. Is this not the case?

crollinson:
@mdietze It looks like within canopy_struct_dynamics.f90 snowfac may be used for both horizontal and vertical uses, but I haven't been able to fully straighten out which is which.

There are definitely places where the horizontal snow_fac is correct such as lines 592-593:
         surf_rough = soil_rough * (1.0 - snowfac_can)                                     &
                    + snow_rough * snowfac_can

However, as I look at it a bit more, it looks like there are places that imply vertical snowfac such as 379-388 and 577-587, but I'm not entirely sure.  It seems like horizontal might work here if we switch around opencan_frac and snowfac, but I haven't given it too much thought since I ended up chasing other directions. 
         !     Find the roughness as the average between the bare ground and vegetated     !
         ! ground, and apply the snow cover to further scale it.  The weighting factors    !
         ! are the fraction of open canopy and the fraction of the canopy buried in snow.  !
         csite%rough(ipa) = snow_rough * snowfac_can                                       &
                          + ( soil_rough           * csite%opencan_frac(ipa)               &
                            + csite%veg_rough(ipa) * (1.0 - csite%opencan_frac(ipa)) )     &
                          * (1.0 - snowfac_can)

There's also this in lines 316-320.  I think this is is used for ed_stars and then roughness is overwritten according to the canopy turbulence scheme, but I haven't been able to figure out which this should really be.  I'm thinking probably vertical, but I'm not sure.
         !----- Calculate the surface roughness inside the canopy. ------------------------!
         csite%rough       (ipa) = soil_rough * (1.0 - snowfac_can)                        &
                                 + snow_rough * snowfac_can
         csite%veg_displace(ipa) = vh2dh * csite%rough(ipa) / vh2vr


Due to my confusion, I decided to set it back to horizontal snowfac, but if we can figure out which parts are supposed to be vertical vs. horizontal, it should be a simple fix.

rgknox:
Hi Christy,  looking forward to testing out the cbr changes.  Another request: Could you also make some clarifications on the ED2IN?  Right now it states that ddmort_const is relevant only to cbr scheme 2, but it seems from intuition that it is relevant only for cbr scheme 1.

crollinson:
whoops, sorry Ryan!  The ED2IN has been corrected.

PR #41, opened 2015-03-24
Upstream 24 mar

crollinson:
Updated from Pull Request #33 to include recent updates to SMP and condense various trial-and-error or reversed commits into 2 major commits.  First commit fixes bugs like missing initializations and a few cosmetic things.  Second commit is the implementation of CBR Scheme and a change to how the running carbon balance is calculated.

Note on Running Carbon Balance (per discussions with @mdietze and @ryankelly-uiuc)
@mdietze said that we should be calculating the 12-month running carbon balance for trees from the actual and potential carbon balance, not weighted-average carbon balance scheme that was currently in place.  I think because of the time step, the (11*cpatch%cbr + cbr_now)/12 was incorrectly calculating the running carbon balance ratio.  The CB variables store carbon balance by month and are the correct place to calculate running balances.

Notes on the new CBR Scheme Feature (again per discussions with @mdietze and @ryankelly-uiuc):
There are multiple possible ways the carbon balance could influence mortality.  Rather than impose a single mortality scheme on the user (as was the case previously), the user must now specify the scheme.  We currently provide 3 options: 

2) Implement CBR_SCHEME (new ED2IN option) where the user must now specify how they want CBR (carbon balance ratio) to be calculated. Currently we have implemented 3 options:
0 -- Single stress. CBR = cb/cb_mlmax
cb_mlmax is the carbon balance in full sun and no moisture limitation

1 -- Co-limitation from light and moisture
CBR_LIGHT = cb/cb_lightmax and CBR_MOIST = cb/cb_moistmax
CBR_LIGHT and CBR_MOIST are then weighted according to DDMORT_CONST

2 -- Leibig’s Law of the Minimum — Limitation from either light or moisture depending on which is lower at a given point in time 
CBR = cb/max(cb_lightmax, cb_moistmax)

According to this new schema, DDMORT only applies to CBR_SCHEME 1.


rgknox:
I tested the merged version in the test suite and I see no strangeness.  I compared with r922 which I flagged as a stable version some time ago.  
*Note that the last couple of revisions, including Dan's stuff and Christy's changes to snow roughness should generate (correctly) different results from the previous version.
**Note also that several runs did not complete or are missing.  These were all related to NERSC cluster queuing and job submission system, I found no actual model crashes.

A report can be found here.

https://commons.lbl.gov/download/attachments/105648775/r953_crmergerapid.pdf?api=v2

Can someone confirm that the pdf is viewable?

apourmok:
Thanks Ryan. I can't access the file despite using my BU credentials to
login. I tried both on Google Chrome and Mozilla Firefox.

On Wed, Apr 1, 2015 at 6:19 PM, Ryan Knox <notifications@github.com> wrote:

> I tested the merged version in the test suite and I see no strangeness. I
> compared with r922 which I flagged as a stable version some time ago.
>
>
> *Note that the last couple of revisions, including Dan's stuff and
> Christy's changes to snow roughness should generate (correctly) different
> results from the previous version. **Note also that several runs did not
> complete or are missing. These were all related to NERSC cluster queuing
> and job submission system, I found no actual model crashes.
>
> A report can be found here.
>
>
> https://commons.lbl.gov/download/attachments/105648775/r953_crmergerapid.pdf?api=v2
>
> Can someone confirm that the pdf is viewable?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/EDmodel/ED2/pull/41#issuecomment-88651662>.
>



-- 
Afshin Pourmokhtarian, Ph.D.
Postdoctoral Research Associate
Dietze Ecological Forecasting Lab
Boston University
Deptartment of Earth & Environment, Rm 130
685 Commonwealth Avenue
Boston, MA 02215

DanielNScott:
I can't access it either

rgknox:
thanks, will work on getting it out soon

PR #49, opened 2015-04-08
OpenMP info and seed_dispersal() labeling fix

DanielNScott:
I changed the "parallel info" initialization banner to also include openmp info, so that folks can verify openmp is behaving as they intend.

Seedling stuff was mis-labeled in seed_dispersal(). In the snippet below seedling_mortality is dimensionless in [0,1], bseeds has units kgC/plant, and nplant is plants/m^2 so "nseedling" is a misapplied naming convention.

Everywhere the routine says "density" it's should say "biomass density," or just "biomass" since by-area or by-plant units are the norm for ED biomass.

```fortran
nseedling = donpatch%nplant(donco) * donpatch%bseeds(donco) &
          * (1.0 - seedling_mortality(donpft))
```

The change in initialization output looks like this:
```
+---------------- MPI parallel info: --------------------+
+  - Machnum  =      0
+  - Machsize =      1
+---------------- OMP parallel info: --------------------+
+  - thread  use:       1
+  - threads max:       1
+  - cpu     use:       1
+  - cpus    max:      24
+  Note: Max vals are for node not sockets.
+--------------------------------------------------------+
```

### DIFF
--- a/ED/src/driver/edmain.f90
+++ b/ED/src/driver/edmain.f90
@@ -10,6 +10,7 @@ program main
   !   destroy processes
   !
 
+  !$ use omp_lib
   implicit none
 
   ! command line arguments
@@ -60,6 +61,15 @@ program main
   character(len=*), parameter :: h="**(main)**"
   character(len=8) :: c0, c1
 
+  ! OMP information
+  integer :: max_threads      !<= omp_get_max_threads()
+  integer :: num_procs        !<= omp_get_num_procs()
+  integer :: thread
+  integer :: cpu
+  integer, dimension(64) :: thread_use
+  integer, dimension(64) :: cpu_use
+  integer, external      :: findmycpu
+
   ! For MPI interface 
   integer :: ierr
   include 'mpif.h'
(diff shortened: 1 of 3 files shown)


DanielNScott:
Also, I don't know why this has "2 commits"; They're listed but obviously looking at the files changed it's all just the stuff I described.

DanielNScott:
Based on these resources:
https://www.cac.cornell.edu/education/Training/IntroMay2010/Hybrid-100520.pdf
http://stackoverflow.com/questions/11959906/openmp-and-numa-relation

And the current full openmp specification:
http://www.openmp.org/mp-documents/OpenMP4.0.0.pdf

I believe the best (only real) way to provide predictable results using openmp on multiple-numa-node per compute-node machines might be job-control numa management flags in the vein you've discovered Ryan.

For SLURM the analogue of the "-cc numa_node" flag appears to be "-cpu_bind ldoms" (that's L) which limits jobs to numa-node-local execution.

Unless anyone wants me to keep investigating, this is satisfactory for me, at least for now.

PR #56, opened 2015-04-18
Changes to the anthropogenic disturbance, mpi-free ED and more

mpaiao:
Main changes:

1.  New dist_type categories, which now distinguishes tree fall, logging, fires, abandonment and forest plantations.   The new meaning for cpatch%dist_type.  1 -- Pasture/cropland; 2 -- Forest plantation; 3 -- Tree fall; 4 -- Fire; 5 -- Abandoned; 6 -- Logging.

2.  Rewrote forestry.f90, this file now simply computes the disturbance rates, and let disturbance.f90 to apply them.  Also, cpoly%disturbance_rates is now (hopefully) consistent with biomass harvesting.   

3. Changed minimum height for reproduction based on BCI measurements (tropical trees only).  Values are based on Wright et al (2005), doi:10.1017/S0266467405002294

4. Changes in minimum reproduction size, so seed_rain works again (it was not working before, at least in the tropics).

5.  Added a new mortality parameter (mort0) to allow shifting the curve.  

6.  All routines that contained mpi code became F90, so now it's possible to compile ED without mpi.  To disable it, go to include.mk.opt and replace PAR_DEFS=-DRAMS_MPI by PAR_DEFS=

7. Changed compilation instructions so it can compile most files with -O3 but uses -O2 for files that would otherwise take days to compile with ifort 13 (namely ed_state_vars.F90 and average_utils.f90).  To do so, set F_LOWO_OPTS with lower optimisation level.  If F_LOWO_OPTS is not set, then it is automatically set to F_OPTS.

8.  Miscellaneous minor bug fixes.

--

PS: I haven't worked with forest plantations yet but once a patch becomes a plantation, there is no simple way out.  Before they could become dist_type=3 through fires or tree fall, however we can't use tree fall at least in the tropics because plantations are monospecific, often with non-native species.  The typical LU scenarios don't even expect forest plantations, so they don't have these transitions.  For the time being, I set that tree fall in forest plantation doesn't change dist_type, and fires switch them to abandoned.  Ideas on how to make this better are very welcome.


rgknox:
Marcos' test report can be found here:

https://github.com/EDmodel/ED2Documents/blob/master/EDTS/388a06b_mlo_aa7dc6b_v1_rapid.pdf

apourmok:
I had couple of problems compiling this version of code which I fixed and they all resulted from removing the lines of code below that was removed from initt/ed_params.f90. Is the only reason to remove them that we don't use them it in the model? I noticed seed_rain has its function and not hard coded anymore as constant.

   !----- Define the grass PFTs -----------------------------------------------------------!
   grass_pft=huge(1)
   grass_pft(1)=1
   grass_pft(2)=5
   grass_pft(3)=12
   grass_pft(4)=13
   grass_pft(5)=14
   grass_pft(6)=15
   grass_pft(7)=16
 
  seed_rain(1:17)         = 0.01

mpaiao:
@apourmok grass_pft was redundant with is_grass, so I only kept is_grass.  

I changed the default seed_rain because the previous value was below the minimum size allowed for new recruits, so it was not effectively doing anything.  In case you don't want seed rain, you could probably set it to zero.   In the future we should probably add an ED2IN option to turn seed rain on/off.

PR #69, opened 2015-05-08
generate_deps.sh update for Mac + NetCDF libraries update

gklarenberg:
The “sed” command in generate_deps.sh fails on Unix, it expects an
extension as an argument. Also changed wording so it does not remove
netcdf.mod from mod_netcdf.mod in dependency.mk.
Changes in the include.mk template are for if you’re using the latest
NetCDF install (separate C and Fortran installations, I have NetCDF
4.3.3.1 for C and NetCDF 4.4.2 for Fortran)

### DIFF
--- a/RAPP/build/bin/generate_deps.sh
+++ b/RAPP/build/bin/generate_deps.sh
@@ -4,6 +4,7 @@ includes="-I ${rapproot}/src/include"
 rappsrc="${rapproot}/src/*/*.f90 ${rapproot}/src/*/*.F90 ${rapproot}/src/*/*.c"
 /bin/rm -f dependency.mk
 ./sfmakedepend.pl ${includes} -f dependency.mk ${rappsrc}
-sed -i s@hdf5.mod@@g dependency.mk
-sed -i s@netcdf.mod@@g dependency.mk
+sed -i .old s@hdf5.mod@@g dependency.mk
+sed -i .old 's@[^_]netcdf.mod@ @g' dependency.mk
 /bin/rm -f dependency.alt dependency.mk.old*
+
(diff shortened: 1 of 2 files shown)


rgknox:
Thanks @gklarenberg  I will do a test on my linux machines and check back in.

rgknox:
@gklarenberg ,  are the tilda's on the "-i" in your pull request preserved?  I am confused with git right now.  In the pull request summary, I see no tildas.  But in the master branch on your fork, I think I see tildas.

robkooper:
@rgknox I fixed the same problem in ED/build/bin but used ~ instead (https://github.com/EDmodel/ED2/blob/master/ED/build/bin/generate_deps.sh#L7). In this case .old is appended instead.

RAPP/build/bin/generate_deps.sh
```diff
@@ -4,6 +4,7 @@ includes="-I ${rapproot}/src/include"
 rappsrc="${rapproot}/src/*/*.f90 ${rapproot}/src/*/*.F90 ${rapproot}/src/*/*.c"
 /bin/rm -f dependency.mk
 ./sfmakedepend.pl ${includes} -f dependency.mk ${rappsrc}
-sed -i s@hdf5.mod@@g dependency.mk
-sed -i s@netcdf.mod@@g dependency.mk
+sed -i .old s@hdf5.mod@@g dependency.mk
+sed -i .old 's@[^_]netcdf.mod@ @g' dependency.mk
```
robkooper: is there supposed to be a space between the two \@ symbols?

robkooper:
Looks good to me.

rgknox:
I see... I didn't realize this was in the RAPP folder.  I wasn't looking closely and thought this was in ED/build/bin.

RAPP/build/bin/generate_deps.sh
gklarenberg: Yeah, netcdf.mod is at the end of a line in dependency.mk a couple of times, and I was worried the line ending or return would be messed up if it was replaced with nothing. So I replaced it with a space. Not sure if it's strictly necessary though.

fabeuw:
@gklarenberg ok on ubuntu I actually had to revert to the older sed command to make it. Have you guys tested this on linux as well?

PR #66, opened 2015-05-05
Bugfix for if statement comparing logical with integer

Viskari:
Hello,

 The existing disturbance.f90 had a bug in it, as due to an apparent typo there was a comparison of a logical and integer. Corrected here.

PR #72, opened 2015-05-14
Bug fix stopping crashes at the change of the year

Viskari:
This commit fixes a bug in ed_state_vars.F90 where a wrong switch caused a possible situation where ED2 would use a term not given values in a calculation.

### DIFF
--- a/ED/src/dynamics/disturbance.f90
+++ b/ED/src/dynamics/disturbance.f90
@@ -784,7 +784,7 @@ module disturbance_utils
                                              / ( real(mypfts) * csite%area(npa) )
                                     !------------------------------------------------------!
                                  else
-                                    if ( include_pft(ipft) == cpatch%pft(1) ) then
+                                    if (ipft == cpatch%pft(1) ) then
                                        !---------------------------------------------------!
                                        !    PFTs match, send all disturbed area to this    !
                                        ! patch.                                            !
--- a/ED/src/dynamics/radiate_driver.f90
+++ b/ED/src/dynamics/radiate_driver.f90
@@ -1,320 +1,346 @@
-!==========================================================================================!
-!==========================================================================================!
-!     This subroutine will control the two-stream radiation scheme.  This is called every  !
-! step, but not every sub-step.                                                            !
-!------------------------------------------------------------------------------------------!
-subroutine radiate_driver(cgrid)
-   use ed_misc_coms          , only : current_time          & ! intent(in)
-                                    , radfrq                & ! intent(in)
-                                    , dtlsm                 ! ! intent(in)
-   use ed_state_vars         , only : edtype                & ! structure
-                                    , polygontype           & ! structure
-                                    , sitetype              & ! structure
-                                    , patchtype             ! ! structure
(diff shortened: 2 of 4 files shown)


apourmok:
@Viskari did you get the error message below when running ED? Since I pulled in the latest changes, the model crashes right after initialization and throws me the error below. I implemented your change in hope that fixes the problem. Will keep you posted.

At line 9522 of file ed_state_vars.F90
Fortran runtime error: Array reference out of bounds, upper bound of dimension 2 of array 'osite' exceeded (1 > 0)

Backtrace for this error:
  + function __ed_state_vars_MOD_copy_sitetype_mask_qmean (0x6F47B2)
    at line 9522 of file ed_state_vars.F90
  + function __ed_state_vars_MOD_copy_sitetype_mask (0x75BCEA)
    at line 8867 of file ed_state_vars.F90
  + function __fuse_fiss_utils_MOD_fuse_patches (0x959016)
    at line 3739 of file fuse_fiss_utils.f90
  + function ed_init_atm_ (0x44A640)
    at line 402 of file ed_init_atm.F90
  + function ed_driver_ (0x43442F)
    at line 222 of file ed_driver.F90
  + in the main program
  + /lib64/libc.so.6(__libc_start_main+0xfd) [0x2ae939485d5d]
** Warning: a core dump was requested, but the core sizelimit
**          is currently zero.

Quit

apourmok:
I applied your changes to my ed_state_vars.F90 but still get the same error. I guess my problem is different than yours.

apourmok:
Thanks @Viskari. This problem is solved.

mdietze:
@Viskari this pull contains a bunch of changes to radiate_driver.f90 and ed_params.f90 in addition to a simple one-line fix to ed_state_vars.f90.   I assume that the last is the only thing you actually want pulled at this time?  In the future you might try pushing small changes like this in separate branches so that you don't send the 'development' work you're doing on RTM at the same time as bug fixes.

Viskari:
I created the radiation push a week ago and pushed it to the mainline as
was agreed as it will be required by project is should be in the mainline.
It was just not accepted before the bug fixes came along.

mdietze:
@Viskari pull request #66 had to be cherry picked to REMOVE the radiation portion of the pull request because you were convolving two issues into one pull request. If there's a radiation feature that is being added that needs to be added as a distinct pull request from these other bug fixes. The radiation pull request needs to describe the changes being made to radiation and we need to be able to evaluate that on it's own.

Viskari:
And again, the radiation pull request was there first and in the line for
almost a week before I pushed through the bug fixes, and thus it is a part
of my current ED2 git. When I uploaded the bugfixes, it added it
automatically to my existing pull request despite my attempts for creating
a new pull request.

mdietze:
@Viskari I don't know what you are talking about, there is no pull request about radiation in the system. Please submit these as separate pull requests.

Viskari:
The radiation pull request is no longer there because when you cherrypicked
the first bug fix, it removed the initial pull request which Git had added
that correction to.

PR #74, opened 2015-05-18
Radiation calculations move from ed_paras to radiative_drivers

Viskari:
Moved radiation term calculations from ed_params to radiative drivers in order to vary values via XMLs.

### DIFF
--- a/ED/src/dynamics/disturbance.f90
+++ b/ED/src/dynamics/disturbance.f90
@@ -784,7 +784,7 @@ module disturbance_utils
                                              / ( real(mypfts) * csite%area(npa) )
                                     !------------------------------------------------------!
                                  else
-                                    if ( include_pft(ipft) == cpatch%pft(1) ) then
+                                    if (ipft == cpatch%pft(1) ) then
                                        !---------------------------------------------------!
                                        !    PFTs match, send all disturbed area to this    !
                                        ! patch.                                            !
--- a/ED/src/dynamics/radiate_driver.f90
+++ b/ED/src/dynamics/radiate_driver.f90
@@ -1,320 +1,346 @@
-!==========================================================================================!
-!==========================================================================================!
-!     This subroutine will control the two-stream radiation scheme.  This is called every  !
-! step, but not every sub-step.                                                            !
-!------------------------------------------------------------------------------------------!
-subroutine radiate_driver(cgrid)
-   use ed_misc_coms          , only : current_time          & ! intent(in)
-                                    , radfrq                & ! intent(in)
-                                    , dtlsm                 ! ! intent(in)
-   use ed_state_vars         , only : edtype                & ! structure
-                                    , polygontype           & ! structure
-                                    , sitetype              & ! structure
-                                    , patchtype             ! ! structure
-   use canopy_radiation_coms , only : par_beam_norm         & ! intent(in)
-                                    , par_diff_norm         & ! intent(in)
-                                    , nir_beam_norm         & ! intent(in)
-                                    , nir_diff_norm         & ! intent(in)
-                                    , cosz_min              & ! intent(in)
-                                    , rshort_twilight_min   ! ! intent(in)
-   use consts_coms           , only : pio180                ! ! intent(in)
-   use grid_coms             , only : nzg                   & ! intent(in)
-                                    , nzs                   ! ! intent(in)
-   implicit none
-   !----- Argument. -----------------------------------------------------------------------!
-   type(edtype)     , target   :: cgrid
-   !----- Local variables. ----------------------------------------------------------------!
-   type(polygontype), pointer  :: cpoly
-   type(sitetype)   , pointer  :: csite
-   type(patchtype)  , pointer  :: cpatch
-   real                        :: rshort_tot
-   integer                     :: maxcohort
-   integer                     :: ipy
-   integer                     :: isi
-   integer                     :: ipa
-   integer                     :: tuco
-   logical                     :: daytime
-   logical                     :: twilight
-   real                        :: hrangl
-   real                        :: sloperad
-   real                        :: aspectrad
-   real                        :: sum_norm
-   !----- External functions. -------------------------------------------------------------!
-   real             , external :: ed_zen
-   !---------------------------------------------------------------------------------------!
+ !==========================================================================================!
+ !==========================================================================================!
+ !     This subroutine will control the two-stream radiation scheme.  This is called every  !
+ ! step, but not every sub-step.                                                            !
+ !------------------------------------------------------------------------------------------!
+ subroutine radiate_driver(cgrid)
+    use ed_misc_coms          , only : current_time          & ! intent(in)
+                                     , radfrq                & ! intent(in)
+                                     , dtlsm                 ! ! intent(in)
+    use ed_state_vars         , only : edtype                & ! structure
+                                     , polygontype           & ! structure
+                                     , sitetype              & ! structure
+                                     , patchtype             ! ! structure
+    use canopy_radiation_coms , only : par_beam_norm         & ! intent(in)
+                                     , par_diff_norm         & ! intent(in)
+                                     , nir_beam_norm         & ! intent(in)
+                                     , nir_diff_norm         & ! intent(in)
+                                     , cosz_min              & ! intent(in)
+                                     , rshort_twilight_min   ! ! intent(in)
+    use consts_coms           , only : pio180                ! ! intent(in)
+    use grid_coms             , only : nzg                   & ! intent(in)
+                                     , nzs                   ! ! intent(in)
+    implicit none
+    !----- Argument. -----------------------------------------------------------------------!
+    type(edtype)     , target   :: cgrid
+    !----- Local variables. ----------------------------------------------------------------!
+    type(polygontype), pointer  :: cpoly
+    type(sitetype)   , pointer  :: csite
+    type(patchtype)  , pointer  :: cpatch
+    real                        :: rshort_tot
+    integer                     :: maxcohort
+    integer                     :: ipy
+    integer                     :: isi
+    integer                     :: ipa
+    integer                     :: tuco
+    logical                     :: daytime
+    logical                     :: twilight
+    real                        :: hrangl
+    real                        :: sloperad
+    real                        :: aspectrad
+    real                        :: sum_norm
+    !----- External functions. -------------------------------------------------------------!
+    real             , external :: ed_zen
+    !---------------------------------------------------------------------------------------!
 
 
-   !----- Check whether it is time to update radiative fluxes and heating rates -----------!
-   if (mod(current_time%time + .001,radfrq) < dtlsm) then
+    !----- Check whether it is time to update radiative fluxes and heating rates -----------!
+    if (mod(current_time%time + .001,radfrq) < dtlsm) then
 
-      !----- Loop over polygons and sites. ------------------------------------------------!
+       !----- Loop over polygons and sites. ------------------------------------------------!
 
-      polyloop: do ipy = 1,cgrid%npolygons
+       polyloop: do ipy = 1,cgrid%npolygons
 
-         !----- Find the solar zenith angle [cosz] -----------------------------------------!
-         cgrid%cosz(ipy) = ed_zen(cgrid%lon(ipy),cgrid%lat(ipy),current_time)
-         !---------------------------------------------------------------------------------!
+          !----- Find the solar zenith angle [cosz] -----------------------------------------!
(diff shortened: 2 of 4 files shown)


rgknox:
Thanks for the submission Toni.  
I don't see any  problems in this pull, but I was wondering if any of them dependent on the hydrologic state?  I guess the question I have is, how often do these parameters change in your current implementation?  If they are still static, could we move the calculation of these parameters to before run-time?  Now that I think of it, in the future it might be a good idea to update these parameters based on snow and water cover fractions.

leaf_scatter_vis(ipft)
wood_scatter_vis(ipft)
leaf_scatter_nir(ipft)
wood_scatter_nir(ipft)
leaf_scatter_tir(ipft)
wood_scatter_tir(ipft)
leaf_backscatter_vis(ipft)
wood_backscatter_vis(ipft)
phi1(ipft)
phi2(ipft)
mu_bar(ipft)

Viskari:
None of the equations I moved were dependent on the hydrologic state and
thus will not change over time.

The before run-time switch seems logical, the reason I put the in radiation
drivers to begin with is because it felt logical and I didn't want to start
messing with deeper plot structure. However, if there are suggestions on a
better location, I'll gladly move them.

mdietze:
Given that @rgknox voiced no objections and suggested that it might be good to make optical parameters dynamical in the future, and that this move facilitates the RTM coupling we're working on, I'm going to go ahead with the pull. FWIW, in the future I think some of these could vary not only with moisture but also with plant nutrient status.

apourmok:
@Viskari pulled the latest changes from mainline and got the error below during compilation:

Error: Unexpected !$OMP END PARALLEL DO statement at (1)
radiate_driver.f90:1061.47:

                csite%par_l_beam_max(ipa)    = radscr(ibuff)%par_v_beam_array(t


Any thoughts?

Viskari:
I've been looking at the piece of code, but don't have any answers at the
moment. It doesn't crash for me and I didn't even touch that part of the
code during the transfer. Besides, your issue seems to be related to the
parallelization according to that error message. I'll do a couple of more
checks from my end.

Viskari:
I checked my geo version of ED2, did the latest pull, confirmed that the
code is what is I originally pushed and ran the installation. I got no
errors and it seems to run without issues.

apourmok:
Thanks @Viskari. Fixed that. When I try to run ED, I got the error below:
Program received signal 8 (SIGFPE): Floating-point exception.

Backtrace for this error:
  + /lib64/libc.so.6(+0x326a0) [0x2b6beaf076a0]
  + function init_can_rad_params_ (0x552260)
    at line 815 of file ed_params.f90
  + function load_ed_ecosystem_params_ (0x550D6A)
    at line 202 of file ed_params.f90
  + function ed_driver_ (0x433AEA)
    at line 73 of file ed_driver.F90
  + in the main program
  + /lib64/libc.so.6(__libc_start_main+0xfd) [0x2b6beaef3d5d]
** Warning: a core dump was requested, but the core sizelimit
**          is currently zero.

Viskari:
I still can't replicate your error. I am trying to check things, but I am
not getting this when running on GEO.

apourmok:
@Viskari  I made a clean branch from master, pulled the latest changes, compiled and tried to run and still get the same error! Are you running it in parallel? 
Could you email me your ED2IN and include.mk.opt file?

crollinson:
Okay, I'll weigh in here too.  Toni's pull request has some bugs that prevent it from being compiled with SMP.  

However, I do not get the same error as Afshin when trying to run the bare-bones single-threaded version when I start a spin from bare ground.

crollinson:
Glancing through the radiation code, it looks like there's an issue with the order things are happening & uninitialized or missing values are trying to be passed to the radiate driver... the first error that gets thrown when trying to compile the current mainline version is:

radiate_driver.f90:382.56:

    leaf_scatter_vis = leaf_reflect_vis + leaf_trans_vis
                                                                                   1
Error: Unexpected assignment statement at (1)

Viskari:
I'm looking in to the matter now, but I am still not personally getting any
error messages during compilation and running despite using the current Git
version, which is making it somewhat difficult for me to track down the
bug. I am not, however, compiling it as the SMP version.

Both leaf_reflect_vis and leaf_trans_vis are defined in ed_params, which it
should read first and when I check the output, I am getting defined values
for both of them.

crollinson:
@Viskari We think this probably has to do with your compiling flags, which according to Afshin, are pretty bare-bones.  Could you try compiling with the follow flags and let us know what happens?
F_OPTS= -g -Wall -W -ffpe-trap=invalid,zero,overflow -Wconversion -fbounds-check -fbacktrace -fdump-core -fopenmp
C_OPTS = -03 -DLITTLE

Conversely, if there's a reason I shouldn't be compiling with those certain flags, that'd be good to know too.

apourmok:
Try to compile with these flags and you will get the errors:

F_OPTS= -g -Wall -W -ffpe-trap=invalid,zero,overflow -Wconversion -fbounds-check -fbacktrace -fdump-core -fopenmp
C_OPTS = -03 -DLITTLE

Viskari:
Those flags do cause my model to crash during compilation, true, but at a
different point than either of you according to your error message and
relating to a part of the code that has not been touched at all.

The issue with flags, and the reason I am running it so bare-boned, is that
the they are often compiler sensitive. So me running the bare version of
the code was both to optimize the runs as well as remove the hassle of
unnecessary flags. To be honest, I am kind of puzzled that your flags work
because you don't have -fno-whole-file, which I found to be really
important due to the information transfer structure.

I'll look in to the flag issue later at more depth.

apourmok:
You can compile the code by removing -fopenmp (which is for SMP) and it compiles but then it throws an error which is dividing by zero. I would say you always need to include at least -ffpe-trap=invalid,zero,overflow so it checks for all these types of errors.

Program received signal 8 (SIGFPE): Floating-point exception.

Backtrace for this error:
  + /lib64/libc.so.6(+0x326a0) [0x2aca322196a0]
  + function init_can_rad_params_ (0x551A7F)
    at line 804 of file ed_params.f90
  + function load_ed_ecosystem_params_ (0x5508B0)
    at line 214 of file ed_params.f90
  + function ed_driver_ (0x43377E)
    at line 73 of file ed_driver.F90
  + in the main program
  + /lib64/libc.so.6(__libc_start_main+0xfd) [0x2aca32205d5d]
** Warning: a core dump was requested, but the core sizelimit
**          is currently zero.

Quit

mpaiao:
@crollinson and @apourmok these options are great for debugging, and they should catch errors earlier in the code.  As @Viskari mentioned, when I compiled with gfortran I had to include option -fno-whole-file, but this may be just different versions of gfortran. 

Once you are all set with the code and no longer developing and debugging, you'd probably want to remove most of these flags so the model runs faster.  I rarely compile ED with gfortran, but if interests these were the options I used last time:

Debugging mode:
```
   F_OPTS= -O0 -ffree-line-length-none -g -fimplicit-none -Wall -finit-real=snan           \
           -ffpe-trap=invalid,zero,overflow,underflow -fcheck=all -frecursive              \
           -fno-whole-file
   C_OPTS= -O0 -DLITTLE  -g
```

Fast mode:
```
   F_OPTS= -O3 -ffree-line-length-none -frecursive -fno-whole-file
   C_OPTS= -O0 -DLITTLE
```

apourmok:
-finit-real=snan, -fcheck=all, -fno-whole-file won't work in gfortran that we use for compiling. I compiled with your flags @mpaiao and removed those 3 that I mentioned and still the compilation crashes right at the same spot.

crollinson:
For the record, here's the full list of compiling flags that I'm currently using and I think Afshin uses as well: 

----------------------------------------------------------------------

#Makefile include include.mk.opt
# Define make (gnu make works best).
MAKE=/usr/bin/make

# libraries.
BASE=$(ED_ROOT)/build/

# HDF 5  Libraries
# ED2 HAS OPTIONAL HDF 5 I/O
# If you wish to use this functionality specify USE_HDF5=1
# and specify the location of the include directory
# library files. Make sure you include the zlib.a location too.

# For 64-bit
HDF5HOME=/project/earth/packages/hdf5-1.6.10
USE_HDF5=1
HDF5_INCS=-I${HDF5HOME}/include
HDF5_LIBS=-L${HDF5HOME}/lib -lhdf5 -lm -lhdf5_fortran -lhdf5 -lhdf5_hl -lz
USE_COLLECTIVE_MPIO=1

# netCDF libraries ---------------------------------------------
# If you have netCDF set USENC=1 and type the lib folder
# at NC_LIBS, with the leading -L (e.g. -L/usr/local/lib).
# If you don't have it, leave USENC=0 and type a dummy
# folder for NC_LIBS (e.g. -L/dev/null or leave it blank)
USENC=0
NC_LIBS=-L/dev/null

# interface ----------------------------------------------------
# This should be 1 unless you are running with -gen-interfaces.
# Interfaces usually make the compilation to crash when the
# -gen-interfaces option are on, so this flag bypass all
# interfaces in the code.
USE_INTERF=1

# MPI_Wtime. ---------------------------------------------------
# If USE_MPIWTIME=1, then it will use MPI libraries to compute
# the wall time (the only double-precision intrinsic).  In case
# you don't have it, leave USE_MPIWTIME=0, in which case it will
# use a simpler, single-precision function.
USE_MPIWTIME=1

# Compile flags ------------------------------------------------
CMACH=PC_LINUX1
F_COMP=mpif90
F_OPTS= -g -Wall -W -ffpe-trap=invalid,zero,overflow -Wconversion -fbounds-check -fbacktrace -fdump-core -fopenmp
C_COMP=mpicc
C_OPTS = -03 -DLITTLE
LOADER=mpif90
LOADER_OPTS=${F_OPTS}
C_LOADER=mpicc
LIBS=
MOD_EXT=mod

# MPI Flags ----------------------------------------------------
MPI_PATH=
PAR_INCS=
PAR_LIBS=
PAR_DEFS=-DRAMS_MPI

# For IBM,HP,SGI,ALPHA,LINUX use these:
ARCHIVE=ar rs

----------------------------------------------------------------------


Granted though, I don't really know what I'm doing with most of those compiling flags since it's a file I inherited along the way and have been too afraid to touch too much.

serbinsh:
@crollinson Thanks!  Real quick, can you edit this and use ```  ``` around the code you put in so Git doesn't do all the crazy formatting?  I am struggling to understand the opt file info

crollinson:
@serbinsh Sorry about that!  I forgot to do a preview before posting and was already fixing it as you commented

serbinsh:
@crollinson No worries and thanks!!  [by the way I did this several times until David Lebauer informed me that my posts were hard to read ;) ]

Viskari:
The thing with flags, though, is that they are mostly sensitive to the
compiler. That is one of the reasons I use the barebones one, because that
can be easily passed from system to system and still work. And in this
regard, despite the debugging flags causing an issue during compilation, as
far as I can tell all the results during the run exist and are reasonable.

What I still struggle with, though, is the fno-whole-file, as I needed to
have that on the geocluster, otherwise I could not get ED2 to compile.

crollinson:
@Viskari regardless of the debates over the best compiling flags, the bottom line is that the current mainline Github code won't compile under the same conditions that it did a few weeks ago. The errors it throws are linked to parts of the radiate_driver.f90 that you modified.  Before changes to the radiate driver were made, things compiled and seemed to be working fine.  @apourmok has tried a bunch of different compiling flags and think it's linked to some incompatibility with the SMP implementation from @rgknox that has been working fine otherwise.

Viskari:
I realize that, but at the same time the problem is that some of the issues
you are having have nothing to do with the changes I did. And for example
when I used the flags you two are using, I get an error flag from a part of
the code that was different from yours and I have not touched in anyway nor
does it use the actual values relevant here. Even the SMP compatibility
with the code, the part of the change I did shouldn't affect that at all as
there was no change in the actual constant values defined and as far as I
saw when checking the in/out terms are all set correctly and when I have it
print out the actual values they are not zero, which the issue Afshin is
reporting when not using the SMP flag.

Another thing that might be happening is a cascading effect, which we have
actually seen before. The bugs in ED2 coming after Marco's changes in last
November, but we just started seeing them crashing ED2 compilation about a
month ago for some reason. This might be something similar. I do not close
out that this is an issue with my changes, but I cannot repeat your problem
and my outputs don't show division by zero.

Viskari:
I went through the code I changed and tinkered around, managing to get
through the bugs you were running in to. There is a larger description in
the pull request.

However, as I also mentioned in the pull request, the run still crashes for
me when using the multiple flags, however the error message points at the
water level equations, which I have not touched in anyway, so I didn't
really go in to that at the moment.

PR #80, opened 2015-05-20
Fixed the bug in lsm_hyd.f90

apourmok:
Fixed the bug in rate constant for lateral redistribution time scale, so now it does not divide it by zero and model runs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -151,3 +151,4 @@ ED/build/bin/configure
 ED/build/rad_param.data
 ED/slurmdbg
 R-utils/read.q.temporary.r
+.Rproj.user
--- a/ED/src/dynamics/disturbance.f90
+++ b/ED/src/dynamics/disturbance.f90
@@ -784,7 +784,7 @@ module disturbance_utils
                                              / ( real(mypfts) * csite%area(npa) )
                                     !------------------------------------------------------!
                                  else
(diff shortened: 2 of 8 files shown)


mdietze:
This calculation is incorrect. cgrid%Te(ipy) is a polygon level variable and here it is being recalculated for every site and then renormalized for every site, but the calculation involves a sum over sites. This calculation should be done ONCE during initialization since none of terms ever change through time

apourmok:
I got the error below if I remove my changes (which is the same error that I got before adding these):
Program received signal 8 (SIGFPE): Floating-point exception.

Backtrace for this error:
  + /lib64/libc.so.6(+0x326a0) [0x2b0775b8b6a0]
  + function calchydrosubsurface_ (0xA44347)
    at line 390 of file lsm_hyd.f90
  + function ed_model_ (0x512C0B)
    at line 489 of file ed_model.F90
  + function ed_driver_ (0x434C99)
    at line 297 of file ed_driver.F90
  + in the main program
  + /lib64/libc.so.6(__libc_start_main+0xfd) [0x2b0775b77d5d]

Since you mentioned cgrid%Te(ipy) should not be here, then the problem is that it never passed to this function and it crashes since it divides it by zero. I will work on it.

Issue #85, opened 2015-06-03
Generating diagnostics for photosynthesis


serbinsh:
Hi @mpaiao and/or @rgknox 

@Viskari and I are attempting to generate some standard photosynthesis "response curves" using some simple diagnostic ED2 runs, in some cases with made up meteorology, but in other cases for a real site but with a simple single cohort single patch runs.  We am doing this to better understand the photosynthetic response of ED2 (not GPP per se but individual leaf/cohort for a "single leaf" and with increasing LAI, i.e. scaled within the canopy), in terms of leaf and canopy phys, to different meteorological drivers as well as VPD and soil moisture.  This is for a manuscript as well as for our general interest in what the formulation within ED2 actually produce, once all the other ins/outs are taken into account.  Of course we could simply implement the code outside of ED2 but I would rather obtain the appropriate output from ED2 so that we know it used the ED2 solver and so that we can examine and diagnose outputs for typical met conditions. 

As such, I am attempting to figure out what outputs we would need for this.  The met drivers are easy enough, however VPD and soil moisture will be a little tricky.  There a several variables that could be of interest (e.g. A_open/close, RUBP, LIGHT, CO2) but most do not produce sensible graphs, e.g. photo versus light/PAR (i.e. light response curve).  What should I be looking for that would represent the photosynthetic rate/assimilation rate for a typical "leaf" for a cohort?  In other words, is there an output in the fast files that I can use for this?

In summary we want to generate 

A vs PAR for a "leaf" and various LAIs
A vs temperature 
A vs Ca cureves
A vs VPD
A vs SWC

If we know what output(s) to look at that we could use to represent photosynthetic rate than we can design some runs to get these results.  Also, if this is possible we could consider creating these as diagnostics that would let us know if the response surfaces make sense or if there is an introduced bug.

mpaiao:
Hi @serbinsh and @Viskari,

I'm not sure why the A_RUBP/LIGHT/CO2 aren't working, last time I checked they looked fine.  Which HDF5 variable and which output file type (-S-, -I-, -Q-) are you using?  I think for you analysis you probably want to use the -I- files, and FMEAN_[variable]_CO, though you could also use the history files as long as you only use instantaneous variables.  

Unless there is some bug, these are the fmean variables I would recommend for this analysis:
PAR – FMEAN_PAR_L_CO / LAI.  This is the PAR absorbed by leaves, but you must divide it by LAI to be m2(leaf)
[...]

PR #91, opened 2015-06-25
New photosynthesis variables

mpaiao:
A few diagnostic variables for photosynthesis following issue #85 :
?mean_a_open: assimilation rate if soil moisture was not constraining photosynthesis (not really "open stomata" because VPD and light may close them)
?mean_a_closed: assimilation rate if stomata were closed
?mean_a_net: assimilation scaled by fsw.

PS:  First time I split my working directory in branches, not sure if it works.  Let me know if the pull request has more stuff than it should (esp. changes in allometry).


mdietze:
@mpaiao  FYI, in addition to the photosynthetic variable changes, this pull request included a ton of changes/additions to various R files as well as a lot of changes to radiate_driver.f90 (too big for github's diff to visualize on the webpage). The R changes don't affect ED itself, and the variable additions look straightforward, but could you clarify what the radiate_driver changes were?

mpaiao:
@mdietze The R scripts are mostly for post-processing.  Some people in Paul's lab used to use these scripts, but I think I'm the only one using it nowadays.  We could probably take them out of ED2 git, and I could create a git version on my account. 

I went through radiate_driv.f90 again on my local machine using Linux's "diff -ib", and there is no real code difference.  Some recent commit added extra spaces in the beginning of every line of the code, and I simply removed them.  I also remove a few obsolete allocate/deallocate statements that have been commented out for a while; I think it's safer to delete them, just to avoid memory leaks in case they were accidentally uncommented.

mdietze:
@mpaiao OK. I'm fine with including the R scripts -- you just specifically asked if the pull request had things in it other than the output variables.

@rgknox do you want to go ahead and merge this? Doesn't change any internal calculations so it shouldn't need the full test suite

serbinsh:
@mdietze @rgknox @mpaiao 

What is the status of this merge?  Are we ready to make it so?

mpaiao:
@serbinsh Yeah, I think it's ready to merge, these are very minor changes to the code.

Issue #92, opened 2015-06-30
Growth/Storage Resp. Sources


DanielNScott:
Growth and storage respiration appear to make up substantial fractions of ecosystem respiration in temperate sites. The image below illustrates respiration composition from a simulation of Harvard Forest spanning June of '10 to January of '13... 

![pso_7_ref_resps](https://cloud.githubusercontent.com/assets/7492269/8441269/050eda3a-1f2c-11e5-98f0-c465a4406eac.jpeg)

I'm of the opinion that storage respiration simply should not exist, but putting that aside for a moment it seems like treating storage and growth respiration as "wood flux" is probably not the best approximation.

If nothing else storage respiration should probably be partitioned into above and below-ground components, presumably on the basis of pft allometry/actual biomass, and I would be inclined to say the same for growth respiration given that we don't have a mechanistic way of accounting for it yet. Even better (in my opinion) would be actually creating bole-respiration, and rolling growth and storage resp into leaf / root / bole resp so that "actual" plant structures' respiration fluxes sum to the total plant respiration...

This may also partially resolve the problem that root respiration and heterotrophic respiration, at HF at least, don't come anywhere near the hypothetical ~40/60 split people talk about:

![r_soil_fractions](https://cloud.githubusercontent.com/assets/7492269/8441865/f173a470-1f2f-11e5-9fdc-fe9b550f41cd.jpg)

Oh yeah, and while we're talking about this, can we please nuke virtual leaf respiration? I guess it's a separate issue but evidence for our ability to do so mostly-side-effect-free is above...

Thoughts?

mdietze:
[...]

PR #95, opened 2015-07-07
Completely removed vleaf respiration from the model.

DanielNScott:
Also, got rid of 2 unused local vars in budget_utils.f90.

Virtual leaf respiration is, as I understand it, vestigial junk in the representation of deciduous hardwood physiology. And as noted in issue #92, it comprises only a tiny fraction of respiration.

Here is the indication that it only affects pfts 9:11 (those with phenology scheme 2)...

From growth_balive.f90:
```fortran
                  !------------------------------------------------------------------------!
                  !     Find the "virtual" leaf respiration.                               !
                  !------------------------------------------------------------------------!
                  cpatch%vleaf_respiration(ico) = (1.0-cpoly%green_leaf_factor(ipft,isi))  &
                                                * salloci * cpatch%balive(ico)             &
                                                * storage_turnover_rate(ipft)              &
                                                * tfact * temp_dep
                  !------------------------------------------------------------------------!
```

From phenology_aux we can see the dependence of green_leaf_factor on phenology:
```fortran
   subroutine prescribed_leaf_state(...)
   ...
  (some stuff)
   ...
      !----- Load the values for each PFT. ------------------------------------------------!
      do pft = 1, n_pft
         select case (phenology(pft))
         case (2)
            green_leaf_factor(pft) = elongf
            leaf_aging_factor(pft) = delay
         case default
            green_leaf_factor(pft) = 1.0
            leaf_aging_factor(pft) = 1.0
         end select
      end do

      return
   end subroutine prescribed_leaf_state
```

and from ed_params.f90 we can see the dependence on the storage_turnover_rate:
```fortran
   storage_turnover_rate(1)       = onethird
   storage_turnover_rate(2)       = onesixth
   storage_turnover_rate(3)       = onesixth
   storage_turnover_rate(4)       = onesixth
   storage_turnover_rate(5)       = 0.00
   storage_turnover_rate(6)       = 0.00 ! 0.25
   storage_turnover_rate(7)       = 0.00 ! 0.25
   storage_turnover_rate(8)       = 0.00 ! 0.25
   storage_turnover_rate(9)       = 0.6243
   storage_turnover_rate(10)      = 0.6243
   storage_turnover_rate(11)      = 0.6243
   storage_turnover_rate(12)      = 0.00 ! 0.25
   storage_turnover_rate(13)      = 0.00 ! 0.25
   storage_turnover_rate(14)      = onethird
   storage_turnover_rate(15)      = onethird
   storage_turnover_rate(16)      = onethird
[...]

PR #98, opened 2015-07-22
Growth resp split up.

DanielNScott:
Commit details below. See issue #92 for more discussion.

1) growth_respiration is now leaf_growth_resp, root_growth_resp, sapa_growth_resp, and
sapb_growth_resp. None of the plant internal carbon dynamics have changed.

2) Flag "growth_resp_scheme" added to ED2IN to indicate old or new flux partitioning.
 - If growth_resp_scheme = 0, all growth resp is "in" sapa_growth_resp which leaves
   aboveground wood to the CAS. This is identical to ED prior to this commit.
 - If growth_resp_scheme = 1, growth resp is divided in proportion to biomass into the above
   fields, which enter the CAS from their "correct" respective locations.

3) ed_var_tables max var count changed from 1600 -> 1800


DanielNScott:
Also, I've got this through the EDTS, but don't yet know where to put the report...

mdietze:
Thanks Dan!

If I could make one suggestion, within growth_balive.f90 could you move the calculation of growth_resp_int to be within Case 1, which is the only place it's really needed? Then Case 0 could just do the traditional calculation without risk of round-off errors or division by 0 issues.  I'm also thinking forward to wanting to add a Case 2 where growth respiration could be made proportional to the growth of each pool rather than it's biomass (which likewise wouldn't use growth_resp_int)

DanielNScott:
Ok, updated growth_balive

DanielNScott:
Just checking in, what's the delay merging this and pull req. #95 ? I'd like to get this stuff in ASAP so I can get rid of my branches, and the longer it sits the more work of merging multiple things will accumulate.

Also, if it would be better, I have a local branch which is a merge of this and #95. Should I push that one instead? I figured we'd want two for the record, since they're distinct items really.

mdietze:
@rgknox I've looked through the code and give this a thumbs up. Could you merge or let Dan know where you want the EDTS posted?

mdietze:
Trying to address the pile of pending pull requests and having just done a few other pulls there's now a conflict on this one.  Could you resolve and update the pull request?

DanielNScott:
Ok, everything is up to date, it should commit with no conflicts.

PR #104, opened 2015-08-04
Bug fixes due to contradictions with merged files.

Viskari:
As the title says. There were some contradictions in the files which I had to address in order to get the newly merged ED2 to compile.

Mostly simple things, but I ended up removing effarea_transp from ed_xml_config.f90 as it claimed that it wasn't defined there.


mdietze:
It looks like fmean_A_open and fmean_A_closed were removed from ed_state_vars.F90.  Was this what was intended, and if so why?

Viskari:
Because they were defined there twice.

PR #106, opened 2015-08-10
Streamlined version of ED2 that only calculates radiative transfer

Viskari:
As the title reads, EDR is a streamlined version of ED2 that calculates the radiative variable values for a canopy for one timestep. This is necessary for our project for approximating new radiative values.


mdietze:
@Viskari the large majority of this doesn't seem related to the RTM. Could you clean up files that you don't need (e.g. *.mod files, everything in Template, etc.) Also, unless you've modified *.f90 files, you should be compiling against the versions in the ED folder rather than copying files to the EDR folder. It will be exceedingly challenging to keep ED and EDR in sync if any changes in one need to be copied to the other.

serbinsh:
@Viskari is this pull request now defunct? That is, I think you replaced it with another long ago....so should we remove this request?

Viskari:
It can be removed.

serbinsh:
@Viskari I want to re-confirm that this PR is defunct and should be deleted

PR #107, opened 2015-08-10
Stripped down version of ED2 for radiative parameter experiments

Viskari:
Only includes the specific model files altered. Otherwise it uses the base model code to compile.


mdietze:
Seems like a good start. For the src files, going forward we might want to think about splitting those files within the ED directory into the parts that EDR needs to modify and the parts that it doesn't. Basically, if we can keep the unnecessarily duplicated code in the RTM folder to an absolute minimum it should make it easier to maintain moving forward. Similarly, we might think about refactoring the RTM code in ED to be called twice, once for each band (VIS vs NIR), rather than looping over bands internally, which would make it easier to support a larger number of bands in the RTM without extra hacks.

PR #88, opened 2015-06-11
Adjusting ED2 output for pecan

Viskari:
-Added keys in ed_state_vars.f90 so that FMEAN_* variables are written in the tower output file.
-Added an output variable for FMEAN_A_OPEN_PY and FMEAN_A_CLOSED_PY
-Created new variables for FMEAN_BDEAD_PY and FMEAN_LAI_PY required by pecan.

Issue #109, opened 2015-08-17
SIGFPE Fault: structural_growth.f90


crollinson:
Received at end of the first month of run (June) while spinning up from bare ground.  Haven't dug into it yet since quite a few things have changed since the spring. Running with GROWTH_RESP_SCHEME = 0 (a new setting from my old runs).

Program received signal 8 (SIGFPE): Floating-point exception.

Backtrace for this error:
  + /lib64/libc.so.6(+0x326a0) [0x2aace46136a0]
  + /lib64/libm.so.6(+0x2b6f7) [0x2aace27146f7]
  + /lib64/libm.so.6(logf+0x11) [0x2aace2719291]
  + function update_vital_rates_ (0xC0CB96)
    at line 1212 of file structural_growth.f90
  + function structural_growth_ (0xBFEE24)
    at line 391 of file structural_growth.f90
  + function vegetation_dynamics_ (0xC6FD58)
    at line 88 of file vegetation_dynamics.f90
  + function ed_model_ (0x514F4C)
    at line 401 of file ed_model.F90
  + function ed_driver_ (0x434CD9)
    at line 297 of file ed_driver.F90
  + in the main program
    at line 285 of file edmain.F90
  + /lib64/libc.so.6(__libc_start_main+0xfd) [0x2aace45ffd5d]
** Warning: a core dump was requested, but the core sizelimit
**          is currently zero.

crollinson:
A potential source of the error from uninitialized variables (found in the compiling output):
ed_init_full_history.F90:3443: warning: 'sidx' may be used uninitialized in this function

pidx is used similarly defined in lines above, but here's where the error is being thrown:
         cpoly%disturbance_memory(   1,sidx,isi) = tmp_dist_memory(1,2,isi)
         cpoly%disturbance_memory(   1,pidx,isi) = tmp_dist_memory(1,3,isi)

anybody know what sidx is?  @mpaiao this is in lines of code marked as something brought in when you made a bunch of changes all the way back in November.

another thing that keeps popping up as an error is from this line (ed_state_vars, line 9088, i is uninitialized):
      allind = (/ (i,i=1,isize) /)
I'm not sure what this means, but I'm guessing in this case it's okay.  maybe?



I'm not convinced either of these are the source of the error, but I figured uninitialized values are always worth looking into.

mpaiao:
[...]

PR #111, opened 2015-08-19
correct cohort initialization

crollinson:
fix accidental zeroing of bdead instead of fmean_bdead in #88 (commit
f1bd1b72e47d3bf0adce8a076ae3be0864db8cd9)


crollinson:
addresses #109

PR #118, opened 2015-08-25
cohort fusion updates

crollinson:
a few minor things to improve stability during cohort fusion when spinning up from bare ground:
1) ditch use of "newn" into fuse_2_cohorts because rounding/precision errors make rnplant + dnplant > 1 and can cause biomass to drift by more than 1%
2) add functionality to import the tolerance levels used in cohort fusion to allow greater user control.  In some bare ground spins, default tolerance levels resulted in very large trees >1 m DBH being fused with very small trees <1 cm DBH

also fixed a typo with dmean_leaf_temp


mdietze:
@crollinson in ed_xml_config, when you add variables to the read, you need to parallel that with a writen later in the file so that the xml written into the output directory is a complete record of the settings used.

Otherwise, looks good to me.

crollinson:
@mdietze Right now only PFT variables are being written into the output record.  Should I go ahead and add everything else from above while I'm at it?

crollinson:
I started going through the xml to copy everything into the write and want to ask about lines 138/139: 

         call getConfigSTRING  ('input_filepath','misc',i,cval,texist)
         if (texist) thsums_database = trim(cval)

Should this really be writing the path for the chilling & GDD file or should this be SFILIN which makes more sense with what's around it?

mdietze:
Good question on thsums. I think my original logic may have been that input_filepath points to everything in the ed_inputs directory in the standard collection of input databased that gets carried around with ED, but I agree that this is unintuitive and in reality we should have separate and explicit variables for SFILIN, THSUMS, as well as the other standard databases (VEG_DATABASE, SOIL_DATABASE, etc).  Of these SFILIN is the highest priority since it is the path most likely needing to be changed automatically by code when doing different runs. That said, PEcAn actually changes this in the ED2IN itself, so it's safe to change these (i.e. I don't think you'll hit any backward compatibility issues)

crollinson:
Updated the xml code so now that everything that could be read in is being written out.  I did confirm that the code compiles & runs properly.  

I should note that I did skip land use since I don't quite understand that set of code or how to make it write out properly in the xml.

mdietze:
Thanks for all that!

PR #134, opened 2015-10-29
Working RTM testing module

Viskari:


mdietze:
These changes are 100% outside the ED2 directory so this code is being accepted without running the test suite.

PR #145, opened 2015-12-07
Fixed xfer sign error mistake in cohorts with no carbon left.

DanielNScott:
Dying cohorts were given extra chances at life... I would say I'm sorry, but... 
Ok, I'm sorry, we need to get rid of them.

Note the tr_[pool] variables are added into their respective pools, hence the sign issue.


DanielNScott:
PS, @mpaiao maybe you want to use your new-found code-powers to pull this Marcos. I would feel sneaky if I just merged my own requests without anyone else looking at them!

PR #148, opened 2016-02-03
travis retry & .gitignore ignores /build/ed_2.1-opt*

DanielNScott:
Travis failed to find packages in commit ac48455, trying different ones.


DanielNScott:
another travis fail

PR #149, opened 2016-02-04
Another try at Travis libraries

mdietze:


mdietze:
@DanielNScott since you were playing with Travis too yesterday, I wanted to point out that I've made some real progress toward getting it to work, but now it's failing during the actual ED build (complaining that it can't find gfortran even though it's installed). Here's the build log https://travis-ci.org/mdietze/ED2/builds/107038008
Any suggestions on what I'm doing wrong? Maybe we need to set up a travis-specific OPT case to get all the libraries right?  @rgknox @mpaiao you guys are good with make.

DanielNScott:
(Accidentally hit close, sorry)

mdietze:
@DanielNScott thanks for the feedback. Looks like it's working now.

Everyone else: The current config on Travis-CI will check if ED2 compiles any time someone submits a pull request, as well as any time the code is updated. Obviously people should be testing their code before they submit it, but it doesn't hurt to have an automated check. Down the line, it's also possible to add integration tests (e.g. a short run) to Travis in addition to just the compilation check. Not to get hopes up too much, the full test suite is too big and involved for Travis, plus it wouldn't spit out all the graphs since Travis just logs the command line.

PR #150, opened 2016-02-06
version specific gfortran in .travis.yml

DanielNScott:
I believe the problem with #149 is that mpif90 specifies the compiler as "gfortran" whereas the actual call may need to specify the version. I've had this issue trying to install on Ubuntu, which I was able to solve by soft-linking gfortran to gfortran-4.9 within the openmpi build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: required
 
 before_install:
     - sudo apt-get clean && sudo apt-get update
-    - sudo apt-get install gcc-4.6 hdf5-tools libmpich1.0-dev libhdf5-mpich-dev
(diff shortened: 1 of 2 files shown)


DanielNScott:
It's a little unfortunate that the travis ci command line interface runs on Ruby... I'm loathe to install another language I'll never use, especially on my already overfull windows machine. 

Anyway I assume you've seen this @mdietze, but for others @rgknox @mpaiao, if you're going to fuss with travis you can check this out: https://github.com/travis-ci/travis.rb#command-line-client

I suppose sooner or later I'll end up installing and using it too...

DanielNScott:
Also, I haven't yet figured out what's up with the travis ci automatic build: I've been waiting for it to start working on this pull request and it's not even telling me that it's going to yet... is this just system latency? Do other people experience weird inconsistent lag when they create pull requests?

PR #151, opened 2016-02-09
Doxygen into mainline

DanielNScott:
This shouldn't have any changes in it except the addition of documentation, so I'm pulling it myself. It's pretty heavy but we decided it would live in the mainline repo at the code meeting.


manfredo89:
Hi @DanielNScott , I was just wondering why are you putting all these html files in the repo.
Those are actually produced by the doxygen utility so there is no need for them to be in the repo.
Shouldn't we restrict the Doc to the Doxyfile and a readme and leave the documentation files to be produced by the users?

DanielNScott:
It's a good question, and one I brought up at the code meeting as well. I had also suggested a few alternatives like keeping the documentation in a separate repo, but folks seemed fine with / most interested in doing this. At that time I had also mentioned the size differences.

In any case, I think ultimately this is the best thing to do, since this way the documentation ships with the code. Generally I don't think it makes sense to tell people that if they want to learn about the model they have to first install Doxygen and DOT, then use it to build the documentation themselves. Instead they can pull from github and start poking around the documentation to learn the structure. 

Additionally, I wouldn't be surprised if there was a way to get GitHub to support actually reading through the documentation online, by say using the .io page to link to the index.html file or something.

PR #154, opened 2016-03-21
Master BAG fix

DanielNScott:
Changes:

1) The call to update_ed_yearly_vars() is moved before the ed_output() call, which ensures that e.g. the 2010 yearly file contains growth during 2009-2010 (output month to output month) rather than e.g. the 2011 file containing this information. If there is some reason this should be the latter, let me know!

2) zero_ed_yearly_vars() was missing/never invoked in the code, resulting in the agb and basal area growth and mortality numbers being cumulative sums over the course of a run rather than yearly sums. This was reflected in these vars' "total_" numbers as well. A call was added after output.

3) Because e.g. basal_area_growth was never zeroed, but is now zeroed on a yearly time step, the factor of 12 in it's calculation (which presumed monthly zeroing) was incorrect. Rather than zeroing the values on a monthly time-scale, the factor has been removed, because the values are not used in any other computations, with the result that the by-month interpretation isn't necessary.

I'll put this through the test suite but wanted to get others' input first.
Thoughts?

### DIFF
--- a/ED/src/driver/ed_model.F90
+++ b/ED/src/driver/ed_model.F90
@@ -437,6 +437,17 @@ subroutine ed_model()
          end if
       end if
       !------------------------------------------------------------------------------------!
+      
+
+      !------------------------------------------------------------------------------------!
+      !      Update the yearly variables.                                                  !
+      !------------------------------------------------------------------------------------!
+      if (analysis_time .and. new_month .and. new_day .and. current_time%month == 6) then
+         do ifm = 1,ngrids
+            call update_ed_yearly_vars(edgrid_g(ifm))
+         end do
+      end if
+      !------------------------------------------------------------------------------------!
 
 
       !------------------------------------------------------------------------------------!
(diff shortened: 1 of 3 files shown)


mdietze:
I don't see anything here that worries me. :+1: for running the test suite, imho

DanielNScott:
Ok, thanks for checking it out Mike. I'll post test results shortly.

DanielNScott:
EDTS Report:

[8c00402-dns-af6fa97-v1_rapid.pdf](https://github.com/EDmodel/ED2/files/195929/8c00402-dns-af6fa97-v1_rapid.pdf)

Looks good to me.

DanielNScott:
Hey @mpaiao would you take a quick look at this and pull it if you deem it ok?

mpaiao:
It looks fine and I merged it, my only minor comment is about the annual update in June (which was already there).  We probably want to give the user the option to set the month according to their POI, e.g. if someone decides to run ED for temperate Southern Hemisphere.

DanielNScott:
@mpaiao Ok, thanks Marcos. I believe that functionality is partially if not fully built in already. I'll check, and if it isn't I'll update it.

PR #175, opened 2016-06-06
Quick update to some variable names in ed_xml_config

fabeit:
Misspellings in treefall_s_gt and fire_s_gt, deleted treefall_gt, I didn't see anywhere that variable in the current code.

### DIFF
--- a/ED/src/io/ed_xml_config.f90
+++ b/ED/src/io/ed_xml_config.f90
@@ -401,7 +401,6 @@ recursive subroutine read_ed_xml_config(filename)
            call getConfigREAL  ('seedling_mortality','pft',i,rval,texist)
            if(texist) seedling_mortality(myPFT) = real(rval)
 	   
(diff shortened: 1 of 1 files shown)


fabeit:
Misspellings in treefall_s_gt and fire_s_gt, deleted treefall_gt, I didn't see anywhere that variable in the current code.

mdietze:
treefall_gt was there for backward compatibility

fabeit:
ok I put that back in, and I added an if clause.

mdietze:
You'll want to delete the line
```
if(texist) treefall_gt(myPFT) = real(rval)
```
because the variable treefall_gt doesn't exist internally in ED (which is causing the build to fail). In addition, the logic you've set up is such that this variable is being set to the value of the variable above it in the code (seedling mortality).

fabeit:
Ok fixed that.

PR #185, opened 2016-10-04
Modularize

rykelly:
This is a branch I worked on briefly when trying to get ED compiled on a new machine. I believe it's a well known issue that some (newer?) compilers don't like the ED code. Hope so, in fact, because I honestly don't remember any of the details. Something about older gfortrans allowing routines to be defined in as many files as you want (with a compiler flag), but new ones need them to be wrapped in modules and `use`d. Anyway, somehow I figured out that this would fix the compile issues I was having, and went ahead and did it. Pretty straightforward, except in a few places it was a little tricky to resolve circular dependencies, etc. 

I moved to a new machine with a different compiler when I was almost done with this, so I finished up and confirmed it compiles, but never really tested. @ashiklom recently took up the effort and seemed to have good results. Hope it helps.

P.S. I have no idea what the protocol is for committing (or not) files in ED/build. I left everything in but I suspect you normally don't...


ashiklom:
I added a few commits to @rykelly 's branch, after which I was able to compile and successfully run this version of ED with gcc 5.3.0 and 6.2.0.

Some results (PEcAn workflows):
* [1000002663](https://test-pecan.bu.edu/pecan_ashiklom/08-finished.php?workflowid=1000002663) -- one year run (2004), from bare ground, at Willow Creek with Ameriflux met.
* [1000002664](https://test-pecan.bu.edu/pecan_ashiklom/08-finished.php?workflowid=1000002664) -- multi-year run (1996-2005), from bare ground, at Willow Creek with Ameriflux met.

serbinsh:
On this, myself and @ashiklom have both been getting these warning messages at run completion. 

```
 === Finish initialization; CPU(sec)=6.82; Wall(sec)=6.00; Time integration starts (ed_master) ===

 === Time integration starts (model) ===
 - Simulating:   07/05/2000 00:00:00 UTC
 - Simulating:   07/06/2000 00:00:00 UTC
 - Simulating:   07/07/2000 00:00:00 UTC
 - Simulating:   07/08/2000 00:00:00 UTC
 - Simulating:   07/09/2000 00:00:00 UTC
 === Time integration ends; Total elapsed time=        7.5  ===
 ------ ED-2.2 execution ends ------
Note: The following floating-point exceptions are signalling: IEEE_DENORMAL

```

My environment is:

```
[sserbin@modex PEcAnRTM-EDR]$ uname -a
Linux modex 2.6.32-642.4.2.el6.x86_64 #1 SMP Tue Aug 23 11:15:56 CDT 2016 x86_64 x86_64 x86_64 GNU/Linux

[sserbin@modex PEcAnRTM-EDR]$ module list
Currently Loaded Modulefiles:
  1) /gfortran/5.4.0         2) /gcc/5.4.0              3) /m4/1.4.17              4) /autoconf/2.69          5) /openmpi/2.0.1-gnu540   6) /hdf5/1.8.17-gcc540

```

rgknox:
I reviewed the changes @rykelly. From what I can tell, the changes make sense to me too.  Am I correct in that where you move subroutines it was for the sole purpose of fixing dependencies?  When everything is converted to a module, build order starts getting more complicated, so various subroutines need to be migrated.
I would recommend making sure the model compiles on the intel processor (wasn't sure which one you used yourself @rykelly, but I see @serbinsh uses gfortran).  
Also, this change-set should generate exact same results as previous runs, right?  Ideally, it would be good to see if your new changes reproduce the same results as a simulation built on the base of this change-set. Is there indication people are getting the same results? 
@mdietze said that this branch has been adopted by numerous people in his lab, so it sounds like it is stable in terms of generating results over a range of use cases.

rykelly:
That's correct, @rgknox — any reorganization was just to fix the circular dependencies that arose from wrapping existing code in modules. Since this was all just reorganization I do think it's true that results should be unaffected, but I have to admit I didn't test this myself. I'm actually not using this branch because I moved to an older branch of ED (for unrelated reasons) and never bothered to merge. If @serbinsh is using it on the regular now though he might be able to speak to that.

ashiklom:
Can confirm that this compiles and generates sensible output using the latest gcc/gfortran (6.2.1), HDF5 (1.10.0-4), and OpenMPI (1.10.4-1) on my personal Arch Linux machine.

Here are results for some runs I did with PEcAn on our BU cluster (gcc 5.3.0) are here:
* https://test-pecan.bu.edu/pecan_ashiklom/08-finished.php?workflowid=1000002663
* https://test-pecan.bu.edu/pecan_ashiklom/08-finished.php?workflowid=1000002664
* https://test-pecan.bu.edu/pecan_ashiklom/08-finished.php?workflowid=1000002665

serbinsh:
I have run this version of ED2 on our server with the listed modules and it produced expected results for a series of ED2+PEcAn runs.

mdietze:
@rgknox I don't think we have access to an intel compiler. Could you help us check that?

fabeit:
@serbinsh Regarding the floating point error, I don't know if it's related to that, but I think we should all add the floating point flag -fp-model source, or the equivalent on gfortran. It makes floating point calculations consistent and at least in my test does not slow down the simulation.

rgknox:
@mdietze I have never run ED2 on an intel personally, I am a gnu user myself.  Any intel users out there?

mdietze:
This PR has been open for months and has been the only version that works for my lab that whole time. All changes don't affect model operations, just the ability to compile on current fortran compilers

PR #186, opened 2016-10-07
Mortality bugfix

ashiklom:
If `negligible_nplant` is 0 for a PFT, the `log` operation in the `monthly_dlnndt` calculation fails (`log(0)`) and causes the entire run to die. This introduces a simple check wherein the adjustment is only performed if `negligible_nplant(ipft)` is zero (to account for rounding errors, I define zero to be less than the smallest real of its type, as defined by `tiny`).

As far as I can tell, there are no edge cases where this would break something that should not have otherwise been broken. But, an alternative is to explicitly throw an error if `negligible_nplant` is zero -- that would be much less cryptic than the current mysterious `SIGFPE`.


ashiklom:
The better solution to this is to create a module that enforces parameter restrictions and fails loudly if any of them are violated.

PR #191, opened 2017-01-24
Just updates to the EDR folder

Viskari:


serbinsh:
Updated EDR code to work with multiple PFT inputs and parameterizations (optics, leaf traits, canopy properties).  This does not impact the base ED2 code

PR #194, opened 2017-04-25
EDR revisions

ashiklom:
Summary of changes:

* Replay Ryan Kelly's module changes onto EDR (see #185) build files.
* Replace EDR files identical to ED counterparts with symbolic links
* Fix some EDR-specific bugs (by @Viskari). 
* Additional `include.mk` files for both ED and EDR, useful on the Boston University shared cluster and on "standard" Linux installations (e.g. Debian-based `apt`, Arch's `pacman`).


mdietze:
@Viskari could you confirm that this supersedes #106 and you can close that PR

serbinsh:
@mdietze We closed #106, we should merge this into mainline.

PR #200, opened 2017-07-11
LIANAS + CLEANING

manfredo89:
This is the first attempt to merge lianas into the ED model. Also during these months I've systematically corrected the code for warnings given at compile time.
Most of them where unused variables or subroutines.
Errors where corrected according to a code coverage report. Some modules that were not covered during my sims like the different photosynthetic drivers (euler/hybrid/heun) and the MPI where not corrected (around 10 files still have warnings).
I've also merged a couple of files (like self contained great_circle).
Also I got rid of ".mk.opt" / ".mk.dbg" since the debug/optimization information is already inside the include. 
I've tested the updated code for a number of different scenarios checking that I would have always the same output as before.


manfredo89:
This is the result of the EDTS.
[r85v2ghubrapid.pdf](https://github.com/EDmodel/ED2/files/1295733/r85v2ghubrapid.pdf)

Intel is now set to be the standard compiler. This can be reverted to gfortran to make Travis CI work, however I also changed the structure of the include.mk. So for every patform we should have only one file that contains the opt/debug informations.
After this pull request simulations might give a segfault when reading in the ED2IN. This is because I removed the NL%IOPTINPT modules. The fix is simply to remove the corresponding line in the ED2IN.

PR #205, opened 2017-09-28
this is a tune for performance. The loop inside rk4_derivs (hot spot) has …

manfredo89:
…been optimized


manfredo89:
Serial runs get a bump of about 20/30% (I'm having mixed results) with this optimized loop. More details about the optimiztion technique can be found in this SO discussion:
https://stackoverflow.com/questions/45437350/optimizing-fortran-code-with-intel-vtune-analyzer

PR #206, opened 2017-09-28
enhanced dev_test.m robustness and updated the make scheme

manfredo89:


ED/build/install.sh
```diff
@@ -122,7 +122,7 @@ cd ${BIN}
 # Link to makefiles, includes, and shell scripts
 ln -sf ../make/*.mk ./
 ln -sf ../make/Makefile ./
-ln -sf ../make/include.mk.${PLATFORM} ./include.mk
+ln -sf ../make/include.${PLATFORM}.mk ./include.mk
```
crollinson: Can you provide rational for why you've changed the naming order/scheme?  This would change how EVERYONE has to do their compiles and so I don't think we should make a change like this without 1) clear documentation for EVERYONE (that means changing ALL of the install file names) and 2) providing a rationale for why this is better than the status quo

ED/src/init/ed_params.f90
```diff
@@ -2483,17 +2483,17 @@ subroutine init_pft_alloc_params()
...
    !---------------------------------------------------------------------------------------!
    !     Wood density.  Currently only tropical PFTs need it.  C3 grass density will be    !
    ! used only for branch area purposes.                                                   !
    !---------------------------------------------------------------------------------------!
    !---- [KIM] new tropical parameters. ---------------------------------------------------!
-   rho(1)     = 0.40   ! 0.20
-   rho(2)     = 0.40   ! 0.53
-   rho(3)     = 0.60   ! 0.71
-   rho(4)     = 0.87   ! 0.90
+   rho(1)     = 0.20   ! 0.40
```
crollinson: When proposing new default parameters, please be more descriptive in your commits about why things are being changed and where the numbers are coming from.

ED/build/install.sh
manfredo89: I changed this because there were two include files for every platform, that is opt and dbg. Some commits ago Dan incorporated the options into the include file (see https://github.com/EDmodel/ED2/blob/master/ED/build/make/include.mk.opt.intel) but the naming convention stayed the same, resulting in two identical copies (like include.mk.opt.intel and include.mk.dbg.intel). Since the options for the optimization are inside the file I thought it would make sense to update the install.sh accordingly.
This change happened in PR #200 where 
ln -sf ../make/include.mk.${OPT}.${PLATFORM} ./include.mk
became
ln -sf ../make/include.mk.${PLATFORM} ./include.mk

This present commit only changes the naming convention putting .mk at the end in accordance to standard practice (https://www.gnu.org/software/make/manual/html_node/Include.html).

Now I understand that this is annoying because the includes have to we updated but I think it's good to avoid multiplication of files that are copies of one another. 
So I suggest that by looking at the example for intel (https://github.com/manfredo89/ED2/blob/0d08d0488269b014cf0ba5231cecb1d840d3d6a8/ED/build/make/include.intel.mk) you incorporate the compiler options inside your platform specific include. 
Also by doing this we could probably later on delete some includes that are obsolete (like the one for macOS does not work, and probably it's not the only one).
As a side note I think that most people should be able to compile with the intel platform which also provide better performance than other compilers. So do you have particular reasons not to use intel compilers?

ED/src/init/ed_params.f90
manfredo89: If you check your repo before PR #200 you will see that what this commit does is only to revert to how the numbers were originally. Around line 2490 there was a comment saying 
`[KIM] new tropical parameters.`
and commented next to the values where the numbers you saw.
So I thought these were more updated parameters. Later on, talking with Marcos, I realized that the new parameters where the one actually being used and I consequently updated the PR #200. Since this present PR was made before amending the change, it appears as if there is a difference (not sure why this is happening though).

mpaiao: @crollinson, I had noticed in the first PR from @manfredo89 that the default values for rho were different from the previous mainline, and I asked him to revert to the original ones.  The numbers in this PR are the original numbers.

ED/build/install.sh
mpaiao: @manfredo89 Yes, intel compilers are not free.

ED/src/init/ed_params.f90
manfredo89: The strange thing is that also the numbers in PR#200 are correct. I guess this appears as a change because my ed_params had the other values in before commit 6a13a18048bb00aa7bdcbe1b52836f767f26c3fb (https://github.com/EDmodel/ED2/pull/200/commits/6a13a18048bb00aa7bdcbe1b52836f767f26c3fb)

ED/build/install.sh
manfredo89: I downloaded them from the intel website for free. You just have to give an academic email address. I am just not sure whether only students qualify or postdoc/professors as well.

mpaiao: According to the [intel website](https://software.intel.com/en-us/qualify-for-free-software/), free licence is only for students and professors aiming to use it in classrooms.  People actively developing ED2 may be able to qualify through the open-source contributor (not sure about this one), but in any case, not everyone using ED2 is developing the code.  As far as I know, gfortran and gcc don't have such restrictions.

crollinson: Just to reiterate what Marcos said, we should absolutely NOT assume everybody will use the same compiler or randomly delete compile scripts.  Having compiled (slightly older) versions of ED on about 7 different systems now, do not underestimate how finicky the build can be.  

Also, I wouldn't get rid of the current MacOS build script until someone gets a chance to update it or consensus is that it will no longer be supported.  In general, unilaterally removing or adding things without heads-up or opportunity for group discussion make tracking down the origins of bugs and breaks much harder.

serbinsh: Just wanted to weigh in as well and support the maintaining support for multiple compilers.  My group has exclusively used gfortran / openMPI for compiling/running ED2 on several different platforms and OSs, as well as HDF5 libs.  I do not want to loose the capacity to be flexible with compilation support since, as far as I know, it does not really add a ton of overhead.  As it is now, different groups maintain their specific opt files to work on specific machines and I don't have a problem with that.  I would say that after that big make / install changes that happened about a year ago the build was working well.

manfredo89: Ok. Let's get this straight then. First I am not advocating to scrap support for compilers. I am perfectly happy with people using gfortran. This commit is about avoiding the duplication of includes (opt/dbg) where most of the rest of the file is the same.

About Mac I have solved the issues and committed the changes some time ago. The problem was that the ld utility was not adding common symbols to the archive. Some background here:
https://stackoverflow.com/questions/33277964/linking-problems-for-non-explicit-interfaces
https://stackoverflow.com/questions/34595766/object-files-not-properly-added-to-archive-on-mac
However I should say that I don't know of anyone else having compiled ED on a mac (there were also issues in the code related to that, see the get_cpu function).
So since I can compile on my mac without the need for a specific include what is then the reason to keep it?

serbinsh: OK, fair enough.  It looks like folks would have to update HDF and compiler options, accordingly but removing duplication is helpful if it works across machines.

manfredo89: Also as a side note, beside adding lianas which is "my stuff", most of the content of the PR #200 was removing unused variable that were giving (hundreds of) warnings at compile time. When compiling with all warnings enabled the compilation fails because of some interface problems. I have already solved this issue and I am now able to compile with all checks enabled. I hope to be able to merge also this last improvements at some point.

The way I see this is as something beneficial for everyone, since for me it's really hard to debug code that gives a thousand warnings. When I started using it it took me almost a month to compile on my mac. This is also because the macosx include was not maintained. 

My view was that what is personal stays on the personal branches and what is common (and maintained) stays on the main branch. I think this is a reasonable approach to open source software. Hope this makes sense.

mpaiao:
I downloaded and compiled Manfredo's pull requests on the cluster I use.  I compiled with both intel and gnu, and both compiled with debugging options and gave no error. However, I had to copy the include.mk.opt to include.mk because the file was missing. 

@manfredo89, perhaps you could make the include files consistent if everyone agrees.  Right now some types (intel, gfortran) have 3 types (include.mk.xxxx, include.mk.opt.xxxx, include.mk.dbg.xxxx), and some don't have include.mk.xxxx, which in your version is  the one that is actually used by install.sh.  My suggestion is that all include files that used to be include.mk.opt.xxxx should become include.mk.xxxx, and those that have different compiler options between include.mk.dbg.xxxx and include.mk.opt.xxxx should be merged using variable KIND_COMP.

manfredo89:
@marcos I agree to merge them and it is indeed what I did with include.intel.mk (https://github.com/manfredo89/ED2/blob/0d08d0488269b014cf0ba5231cecb1d840d3d6a8/ED/build/make/include.intel.mk). Note also that the name is include.intel.mk instead of include.mk.intel, motivation is a few posts above.

To me what would make the most sense would be to have one file where we specify the locations and eventually system specific commands and then a few includes which cover for most system / compilations types.
For example I think the location of HDF5 libs should be in a separate, branch/user specific file. While the compilation options (for gfortran and intel) should appear just once each. 
Again rationale for this is that even for compilation a lot of flags are either obsolete or unnecessary like ftrapuv and many others creating overhead and sometimes mistakes.

I think that the include in my development branch (https://github.com/manfredo89/ED2/blob/redux/ED/build/make/include.intel.mk) has eveything necessary for the intel platform (@mpaiao in that branch with all warnings enabled there is no need for 2nd compilation to detect interface problems) and nothing more.

Issue #210, opened 2017-10-06
Failing Build: Function 'machs' at (1) has no IMPLICIT type


crollinson:
I'm fast forwarding a lot to the latest version of ED on GitHub and I'm unable to build.  I'm getting a few variables about missing type.  It looks like this goes back to #200 by @manfredo89 since when I add some calls back in, it gets further.  (see commit cb30c21985f440a45503373c76b66a9b1e4ccf11 "major revision..."

I'll keep debugging since it looks like this isn't the only place with issues, but I want to see if I'm the only one that has had this fail... I see the travis build has been failing lately. 

Note for @manfredo89, it's very hard to see if this is where things came in because the commit that I think this links back to had  555 changed files and too many differences for the git diff or "blame" functions to work.  In the future, please make these smaller commits to make them easier to track.  

Also, please make sure all compile checks from a clean build work before pulling into mainstream since I'd be surprised if I'm the only one getting this error.

manfredo89:
Hi @crollinson, I would suggest you try to incorporate the next pull requests (#205 #206 ) and try to recompile. I've recently been in contact with @mpaiao about merging these two small updates and I think it should soon be feasible. I'm saying this because I've modified the way install.sh calls the make (this is also the reason for Travis failing to build).
However this should not be the cause of missing types.
I am aware of a bug in file ed_read_ed10_20.f90:907 where it should now be
`         call init_ed_site_vars(cpoly)
`
instead of
`         call init_ed_site_vars(cpoly,cgrid%lat(ipy))`
though this was not producing errors with the --kind A compilation on intel platform (with the flags we were using).
If the error you are getting is different please post the code, I will have a look at it.
Also regarding the size of the PR there were a couple of issues. Some files appear to have changed even though they were not. This was due partly to automatic change of permissions. I will try in the future to be more efficient but I think it would really help if there were some guidelines on the git workflow to update the code.

manfredo89:
Hi @crollinson did you manage to compile?
Also be aware that the optimization entry in the ED2IN was removed so you have to delete the following lines

```

   !---------------------------------------------------------------------------------------!
   ! IOPTINPT -- Optimization configuration. (Currently not used)                          !
   !---------------------------------------------------------------------------------------!
   NL%IOPTINPT = ''
   !---------------------------------------------------------------------------------------!

```

crollinson:
I haven't not made progress on compiling ED (out sick and then have other priorities at the moment).

[...]

PR #211, opened 2017-10-10
Fix MPI compiling fail

crollinson:
Per Issue #210 

Not the full fix with the MPI-specific flags, but it allows non-intel MPI compiling to work.

Also a little tweak to allow >1000 yrs of land use change drivers.


manfredo89:
I think most changes are ok. Though I don't think we should reintroduce ioptinpt. As it is now the code would not compile since the file memory/optimiz_coms.f90 where the module is declared is no longer. I can see that it is called by MPI_Bcast but since the variable ioptinpt was not set (was '' in ED2IN) I don't see how it could have been of any use.
I think the easy fix is simply to delete this two calls, however I am not an expert on MPI (my understanding is that ioptinpt was just broadcasted to the nodes, so the calls can be deleted).

mpaiao:
I agree, the MPI_Bcast calls that have ioptinpt (two of them) should be removed in ed_mpass_init.F90.

crollinson:
I do get this error, but the program does seem to compile and run at least with very rudimentary tests `*** No rule to make target 'optimiz_coms.mod', needed by 'ed_mpass_init.o'.  Stop.`

mpaiao:
I can think of two possibilities...
1. You had some optimiz_coms file from a prior compilation
2. There are some calls to optimiz_coms in rules.mk 

Does the problem show even when you compile from scratch?

crollinson:
My guess is there's a optimized_coms in rules.mk since I always do clean installs

PR #212, opened 2017-10-11
Additional fixes for MPI, gfortran and Travis CI

mpaiao:
This review is a follow-up of the latest commits.  I won't merge because I am sending the pull request, @mdietze, @DanielNScott or @rgknox, could one of you please check this PR and merge/suggest edits? Thanks.

Summary of changes (all minor):

1. Some files still had MPI variables that were missing. Also, all variables added back in this commit and in #211 were put inside preprocessor instructions, so they will be skipped in case MPI is not enabled, and included for mpi-compatible compilations.

2. Clean-up of the ED/build directory.  
a. A few minor fixes in install.sh, and I added back the capability of full interface check (intel only).
b. Merged "opt" and "dbg" files into single files.  Those that had differences were merged and the options are now defined through variable KIND_COMP (check install.sh for more information).

3. Travis CI is working again.
a. I added a include.mk.travisci with Travis CI-specific instructions
b. Replaced version-specific packages with generic ones (some versions were no longer available).
c. Set openmpi as the default. Possibly my fault, but I was unable to finish compilation with mpich.

@manfredo89 I couldn't find you through the review tag, could you please take a look at it too?


manfredo89:
Looks fine to me.
I still think that the notation
include.${PLATFORM}.mk
would make more sense than 
include.mk.${PLATFORM}
but that's maybe a personal view (and a tiny detail).

crollinson:
Thanks for the clean-up @mpaiao. It compiles and executes for me with MPI using a Linux machine.

I think making gfortran the default compiler makes the most sense given my experience with several different machines/HPCs.

Personally, I'd rather stick with include.mk.${PLATFORM} since that's most similar to what's been standard in ED for quite some time.

So looks good to go to me and it's nice to see it passing the travis build again.

ED/build/make/include.mk.arch_linux
```diff
@@ -41,29 +41,40 @@ USE_INTERF=1
...
-
 LOADER=mpif90
+#F_COMP = gfortran
+#C_COMP = gcc
 #LOADER = gfortran
 
-LOADER_OPTS=${F_OPTS}
+##################################### COMPILER OPTIONS #####################################
+#------------------------------------------------------------------------------------------#
+# A/B/C/D. Debugging, strictest compilation flags, lowest performance.                     #
```
crollinson: Probably should add a note that these options compile with mpi turned on while the E doesn't.

ED/build/make/include.mk.bu_geo
```diff
@@ -75,17 +75,31 @@ USE_MPIWTIME=1
...
+ifeq ($(KIND_COMP),)
+   KIND_COMP=E
+endif
+ifeq ($(KIND_COMP),$(filter $(KIND_COMP), A B C D))
+   F_OPTS= -g -fopenmp -O0 -Wall -W -ffpe-trap=invalid,zero,overflow -Wconversion -fbounds-check -fbacktrace -fdump-core
+   C_OPTS = -g -fopenmp -O0 -ffpe-trap=invalid,zero,overflow -fbounds-check
+   LOADER_OPTS=${F_OPTS}
+endif
+ifeq ($(KIND_COMP),E)
+   F_OPTS= -g -Wall -W -ffpe-trap=invalid,zero,overflow -Wconversion -fbounds-check -fbacktrace -fdump-core
```
crollinson: you're missing -fopenmp here, but you do have in in the C_opts... should be in both places I think

crollinson: I see this is a holdover from the original make script... pretty sure for MPI compile it needs to be specified in the F_OPTS though.

mpaiao:
@manfredo89 @crollinson Either way is fine with me, to be honest I didn't notice that Manfredo had swapped the order.  In any case, once we agree on the order, some of the file names must be renamed so all include files are consistent.

ED/build/make/include.mk.arch_linux
mpaiao: I simply followed what was already in the original files, but I'm wondering whether these differences are intentional, or just inconsistencies because the dbg was not up-to-date. In case it's the former, I can add the comment.

ED/build/make/include.mk.bu_geo
mpaiao: OK, I will add the option in F_OPTS

rgknox:
I only reviewed the fortran code changes.  I'm very rusty on this code at best, but I think the changes were consistent with intent, that ierr and some other MPI diagnostic/return values are only to be turned on when MPI is active.

mpaiao:
Minor changes following up the discussion. 

1. For the time being we will stick with include.mk.${PLATFORM}.  I renamed include.intel.mk to include.mk.ugent, to make sure @manfredo89's settings are not overwritten, and because there was already a file named include.mk.intel (not sure if anyone is using the latter, though). 

2. include.mk.arch_linux.  I assumed all options should include -fopenmp.  Let me know if this is a problem.

ED/build/make/include.mk.arch_linux
crollinson: I suspect it's just inconsistencies because one or the other was not up to date, but I'm not sure.

PR #230, opened 2018-02-01
New empirical equation for soil moisture limitation

istfer:
## Description

Added a new case for soil respiration limitation due to soil moisture. Updated `decomp_scheme` case selections in other places as well to allow combinations of this implementation together with different temperature limitation etc.

I also updated description in `decomp_coms` and the checks in `ed_opspec`. Let me know if there are other places I need to make changes. I also tested whether it compiles and runs successfully.

## Motivation and Context

@jhmatthes implemented an empirical function a while ago from [Moyano et al., 2012 Biogeosciences](https://www.biogeosciences.net/9/1173/2012/bg-9-1173-2012.html) for soil moisture limitation to soil respiration. I have used it in my runs, I believe @crollinson also tested it a bit, it behaves just fine. Per our discussion during the ED2 meeting I'm pushing it into the mainline.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

## Testing :

- [X] All new and existing tests passed.


crollinson:
The implementation of this looks in line with what Jackie & I had implemented, but I think I remember changes to the definitions of the decay rates for Moyano in ed_params as well.  I can look into it to help, but there's a more pressing concern that needs to be addressed first:  

This has changes that need to be in the ED2IN.  This lead me to discover that **the ED2IN apparently got deleted from the repository in #200 (@manfredo89).**  It should be in ED/run/ and while I have one in my local clone, it's in the online repository. (probably part of a merge conflict where my branch diverges with the master.)  

@istfer: We need the ED2IN back in the repository, so probably best if you do it here, but we'll need to make sure any changes to options get added back to the main ED2IN file.  

@manfredo89, when it gets added back in with the new decamp scheme options, please review it to make sure any of your modifications that require tweaks to the ED2IN file get updated as well.

istfer:
> I think I remember changes to the definitions of the decay rates for Moyano in ed_params as well. I can look into it to help

@crollinson that would be great

I was also looking for the ED2IN file to update the description, I thought that was deleted on purpose. Now committed the deleted version back with updated descriptions, let me know if there are anything else that needs to be changed.

manfredo89:
Hi @crollinson and @istfer. Yes there is a minor change that would need to be incorporated.
At the end of the ED2IN there is a section about an optimization routine that was unused so it's no longer supported. I can see that in the ED2IN that you pushed those lines are commented. Since in the code there is no reference to those I would actually delete them. 
The lines are:
```
   ! IOPTINPT -- Optimization configuration. (Currently not used)                          !
   !NL%IOPTINPT = ''
```

manfredo89:
I deleted those lines my self, if you are ok with the change you can merge the PR

istfer:
yeah it killed my run right away so I commented it out, your deletion is fine with me

crollinson:
So it looks like the one change Jackie made to the ed_params that is consistent with the current ED2 version was [here](https://github.com/jhmatthes/ED2_Erelax/blob/206f380b29bd1d4e49af0bc4e4ad4fe430504029/ED/src/init/ed_params.f90#L1746-L1747)

```
!   resp_temperature_increase = 0.0757
   resp_temperature_increase = 0.23503
```

However, resp_temperature_increase isn't explicitly part of the Moyano equation, so perhaps we should just leave it alone for now?

(While I did test this with PalEON somewhere along the line, none of my final runs used the Moyano equation as I was finding to many other bugs elsewhere in the code and I decided to revert to being based off of the mainline while debugging.)

crollinson:
@istfer before I approve the changes, can you confirm that the ED2IN executes with @manfredo89's updates?  Just want to make sure we catch any overlooked variables that breaks things.

istfer:
@crollinson yes, the model compiles and runs (and finishes) successfully with these changes. The  ED2IN in this PR contains the same variables with the one I used in that run, but of course I populated it differently (through PEcAn) for my run.

PR #234, opened 2018-02-13
Adding plant hydraulics

xiangtaoxu:
This pull request (**hydro-3c03ec6**) adds a few new features from Xu et al. 2016 New Phytologist [X16] to the current mainline master (**e25e9a8**). This is an update for PR #234.

## Description
1. **New Features**:
1.1 A new plant hydro scheme that can track changes in leaf/sapwood water potential, and sap flow. A new ED2IN flag (**PLANT_HYDRO_SCHEME**) has been added in the ED2IN template under run folder. There are four options running with plant hydrodynamics: 1. run hydrualics with hydraulic parameters from Christofferson et al. 2016 GMD [C16], 2. run hydraulics with part of parameters from Xu et al. 2016 New Phytologist, -1 and -2 are the same as 1, 2 except that the model does not track leaf/wood heat capacity changes when leaf internal water changes.
1.2 A new stomatal conductance scheme. One ED2IN flag, **ISTOMATA_SCHEME** is added to ED2IN. 0 -> existing Farquhar-Leuning scheme. 1-> new Farquhar-Katul scheme (only works for C3 plants for now, see X16 as reference).
1.3 Two new plant water stress scheme. Two new options are added for **H2O_PLANT_LIM** in ED2IN: 3. Use leaf water potential to modify fsw following @tompowell9's work. 4. Use leaf water potential to modify Katul's optimization-based stomatal conductance model based on X16.
1.4 A new hydraulics-driven drought-phenology scheme (**IPHEN_SCHEME = 4**), based on X16.
1.5 A new structrural growth scheme (**ISTRUCT_GROWTH_SCHEME**), which reserve enough bstorage to fully reflush the whole canopy and fine roots once before spending carbon in structural growth. This helps to give deciduous PFTs enough carbon to flush new leaves when growing season comes.
1.6 A new mode to consider within canopoy trait pheno-plasticity (SLA and Vm0) based on Lloyld et al. 2010 Biogeosciences. This can be set through TRAIT_PLASTICITY_SCHEME in ED2IN.

2. **Code Organization**
2.1 The plant hydrodynamics calculation is encapsulated in dynamics/plant_hydro.f90. Plant internal water flow rate is assumed to be constant within each model timestep. In order to track changes in leaf internal water during integration, two new variables (leaf_water_int and wood_water_int) are added to rk4type. Changes in leaf energy and heat capacity associated with plant hydraulics are updated accordingly in rk4_derivs.f90 (energy) and rk4_misc.f90 (update_diagnostic_vars, heat capacity).
2.2. To support plant hydrodynamics calculation, I add 9 key cohort-level variables for plant hydraulics into ed_state_vars.f90 as well as corresponding temporal averages. All variables are annotated following Doxygen stile in ed_state_vars.f90. I also add a new variable format (nzg by ncohort, id is 42) for one of the new state variable (**wflux_gw_layer**, root water uptake from each soil layer). About 20 new PFT-specific parameters are added in init/ed_params.f90. These new parameters are all organized into one new subroutine init_pft_hydro_params. However, over half of the new parameters are auxiliary variables used to generate the key hydraulic parameters. The most important/sensitive parameters are xylem hydraulic properties and capacitance (About **6** parameters), which can be obtained from field observations. Two sets of parameterization is provided in the code, one based on Xu et al. 2016 New Phytologist and one based on Christoffersen et al. 2016 GMD.
2.3. Add a new allometric function dbh2sf (dbh to sapwood fraction)
2.4. Modify the calc_veg_hcap in utils/ed_therm_lib.f90 to allow leaf/wood heat capacity change with tissue relative water content. These can lead to up to 50% (or even higher for PFTs with poorly constrained parameters) reduction in leaf heat capacity during dry season/drought.
2.5 The new module is annotated following Doxygen style. I also adapt the old annotation to Doxygen style for some of the existing modules.
2.6 Moved copy_patch_init after canopy_photosynthesis and soil_respiration in the integration schemes. This can lead to issue #236 . To resolve this, a quality check is added in canopy_photosynthesis to initiate leaf_gbw when necessary.

3. **To-Do List**
3.1. Modify EDTS ED2IN templates
3.2. Other code clean-up such as making sure using 'only' when importing modules.

## Collaborators
@tompowell9 @dmedvigy 

## Motivation and Context
This new module provide predictions of PFT-specific plant hydrodynamics, allowing tests over more mechanistic plant water stress functions (stomata, phenology, mortality) in ED2. The new module can also simulate leaf water potential, and sapflow, allowing for comparison and assimilation of corresponding field observations.

## Types of changes

## Checklist:

Should I generate a new Documentation using Doxygen? Where should I put it? I put it in [my personal github page](https://xiangtaoxu.github.io/ED2Doc/index.html) now.

## Testing :

I have run the standard test suites using OMP multi-threading. I have also run several sites using serial setup. The only difference is that some sites crash under debugging mode when running with OMP.
[e25e9a8-XX-3c03ec6-serial_medium.pdf](https://github.com/EDmodel/ED2/files/1764461/e25e9a8-XX-3c03ec6-serial_medium.pdf)
[e25e9a8-XX-3c03ec6-omp_medium.pdf](https://github.com/EDmodel/ED2/files/1764462/e25e9a8-XX-3c03ec6-omp_medium.pdf)


xiangtao-princeton:
@manfredo89  Maybe you can ask Felicien to help reviewing the hydro code as well? I have modified the codes and it should work well with small cohorts now.

manfredo89:
Sure I will try to go through it with him. This new PR sounds really exciting, glad it's making it to the main branch.

xiangtaoxu:
Here is the new test reports including all the test sites except M34 and PRG. For M34, both the mainline version and the test version crashed at the same time due to RK4 underflow. For PRG, input specific humidity is off.

For most of the runs, the TEST and MAIN are exactly the same. However, soil moisture has a significant bias for Toolik runs (TL2). The regional runs at RJG also show some level of relative bias in biomass. I checked the input ED2IN, the input parameters are exactly the same. @mpaiao Do you have any ideas why that happens?

[a7ce6bd-XX-1ba184c-v1_medium.pdf](https://github.com/EDmodel/ED2/files/1725624/a7ce6bd-XX-1ba184c-v1_medium.pdf)

ED/src/dynamics/photosyn_driv.f90
```diff
@@ -14,7 +14,8 @@ subroutine canopy_photosynthesis(csite,cmet,mzg,ipa,ntext_soil
    use pft_coms       , only : water_conductance  & ! intent(in)
                              , include_pft        & ! intent(in)
                              , vm0                & ! intent(in)
-                             , leaf_turnover_rate ! ! intent(in)
+                             , leaf_turnover_rate & ! intent(in)
+                             , leaf_psi_tlp       ! ! intent(in)
```
manfredo89: This will probably gibe a warning of unused variable

xiangtaoxu: Hi, Manfredo

Thanks for your review. This leaf_psi_tlp was used for test purposed and is supposed to be used in my next pull request. I will probably update this pull request with my other new updates (drought phenology, leaf water potential modified stomatal conductance). And perform another round of test suites.

As for M34, I have asked Marcos, he said he encountered this problem before and changes the RK4 tolerance to avoid the RK4 Underflow error. But I'm not sure which modification leads to the error.

manfredo89: Ok. I think it's definitely no prob to update with the next PR. Also I think once this is done it would be great to have part of the test to cover your new features.
To be honest I was also planning to add PFT 17 to the test for at least the GYF site but I didn't have time yet.

xiangtaoxu: Yeah, I'm planning to add Costa Rican sites as additional test suites. Maybe we can coordinate and add two test sites into the suites.

Xiangtao

xiangtaoxu:
@manfredo89  Thanks for reviewing the codes. I think it is better to wait until I figure out why the TEST and MAIN show differences in some of the test sites.

ED/src/dynamics/photosyn_driv.f90
manfredo89: Yes, by the way, I think the Test suite itself also needs some upgrade. I rewrote the script (https://github.com/manfredo89/ED2/blob/redux/EDTS/exec_test_carver_serial.sh and https://github.com/manfredo89/ED2/blob/redux/EDTS/dev_test.m) that creates the qsub jobs so that we need only one ED2IN template per site (so as to avoid the risk of having different parameters for DBUG, TEST and MAIN). Hopefully I'll find time to push it to the main branch.

mpaiao:
@xiangtaoxu It is a bit strange indeed, but I don’t know what could cause this.  I think Toolik uses xml to initialize many parameters, perhaps it’s something defined there that could make the difference.  By the way, I haven’t had a chance to go through the commit yet, will do tomorrow or over the weekend.

xiangtaoxu:
@mpaiao I will try to update the pull request today or tomorrow including the coupling of plant hydraulics and other physiology process and conduct another round of test suites. Hopefully, I can figure out the reasons of the mismatch during the process.

mpaiao:
I saw that @manfredo89 made a comment about unused variables.   I've cleaned my branch by turning all warnings into errors. In ifort, you can add the following instructions:
```Makefile
-warn unused -warn uncalled -warn usage -diag-error=all
```

Because of the differences in compilation instructions when using mpif90 and ifort with PAR_DEFS unset, it's a good idea to make the test with both settings.

manfredo89:
You can also just set `-warn` which activate all warnings, though if I remember correct the compilation fails this way because of interface problems. I think in my branch I solved this by enclosing everything into modules (for some reason I also had to move `new_patch_sfc_props` into `fuse_fiss`).

mpaiao:
@manfredo89 I turned many f90 files into modules, and now ifort 17 compiles with the interface check — something we want to be able to check, as it spots inconsistencies in arguments between call and declaration of functions and subroutines.  This will be in my pull requests, but I am going to wait until Xiangtao finishes his before I start mine.

ED/src/dynamics/disturbance.f90
```diff
@@ -3124,7 +3125,8 @@ subroutine prune_lianas(csite, np, lsl)
                                 , h2dbh                    & ! function
                                 , size2bl                  & ! function
                                 , dbh2bd                   & ! function
-                                , dbh2krdepth              ! ! function
+                                , dbh2krdepth              & ! function
+                                , dbh2sf                   ! ! function
```
mpaiao: I see that you added dbh2sf in the "use allometry, only: ..." blocks in several subroutines, but they don't seem to be used.

ED/src/dynamics/euler_driver.f90
```diff
@@ -9,6 +9,7 @@ subroutine euler_timestep(cgrid)
    use rk4_integ_utils
    use soil_respiration_module
    use photosyn_driv
+   use plant_hydro
```
mpaiao: I know you weren't the one who started this, but as a general comment, use statements should always have the "only" list for clarity.

ED/src/dynamics/rk4_derivs.F90
```diff
@@ -1821,6 +1896,13 @@ subroutine canopy_derivs_two(mzg,initp,dinitp,csite,ipa,hflxsc,wflxsc,qwflxsc,hf
          !---------------------------------------------------------------------------------!
          dinitp%wood_energy(ico) = 0.d0
          dinitp%wood_water(ico)  = 0.d0
+         dinitp%wood_hcap(ico)   = 0.d0
```
mpaiao: We discussed this offline, and this won't affect the legacy simulations.  But just to keep it registered, heat capacity should not be a prognostic variable, ideally internal water mass should be the tracked variable with the differential equation, then heat capacity could be updated diagnostically.

ED/src/dynamics/rk4_driver.F90
```diff
@@ -760,6 +770,51 @@ subroutine initp2modelp(hdid,initp,csite,ipa,nighttime,wbudget_loss2atm,ebudget_
...
       !------------------------------------------------------------------------------------!
       do ico = 1,cpatch%ncohorts
+         !---------------------------------------------------------------------------------!
+         ! First, update variables related with plant hydrodynamics because it can 
+         ! change leaf/wood heat capacity, which will be used later 
+         !---------------------------------------------------------------------------------!
+         if (plant_hydro_scheme > 0) then
+             ! Need to update leaf_water_int and wood_water_int diagnostically
+             ! from changes in heat capacity
+             cpatch%leaf_water_int(ico) = cpatch%leaf_water_int(ico)                    &
```
mpaiao: Same as my comment in rk4_derivs.F90.  Mass should be the prognostic variable, not heat capacity.

ED/src/init/ed_params.f90
```diff
@@ -2443,7 +2452,12 @@ subroutine init_pft_alloc_params()
    rho(3)     = 0.71   ! 0.60
    rho(4)     = 0.90   ! 0.87
    rho(5)     = 0.20   ! Copied from C4 grass
-   rho(6:11)  = 0.00   ! Currently not used
```
mpaiao: I like the idea of assigning actual numbers for wood density for temperate species for completion.  I think it's no longer the case, but if I remember correctly, in the past rho=0 was a special flag to tell not to use wood density.

ED/src/init/ed_params.f90
```diff
@@ -3063,6 +3110,38 @@ subroutine init_pft_alloc_params()
    b2WAI(17)    = 2.0947       ! Broadleaf
    !---------------------------------------------------------------------------------------!
 
+   !---------------------------------------------------------------------------------------!
```
mpaiao: Re-defining sapwood biomass is an excellent idea because as @yjkim1028 had noted a long time ago, the original definition was incorrect (by a lot).  I do have a few concerns on this:

1. If you changed sapwood biomass, then the parameters for bdead must change too, otherwise the size-AGB allometry will be off.
2. For legacy I think this should be optional.  My suggestion to ensure back-compability is that you define this as a separate functional form.  I don't think anyone is using IALLOM=1 right now so you could claim that one unless someone says something.  Otherwise, just set a new IALLOM=4 case.

xiangtaoxu:
@femeunier  The current hydro branch is out of date and still has some bugs. I will push the newest version later today, which also addressed some issues I have discussed with @mpaiao

xiangtaoxu:
@mpaiao Thanks for your review. Please wait until I update my hydro branch. The new version incorporates new drought-driven phenology, optimization-based stomatal conductance, and fixed some bugs as well some issues we discussed about last time.

I have conducted two sets of test suites for this up-coming version (one omp, one serial) and the model can reproduce the major results of my previous runs at Palo Verde, Costa Rica (the inconsistencies are probably due to changes in radiative transfer scheme. I used ICANRAD = 0, which is not used in current master).

ED/src/init/ed_params.f90
xiangtaoxu: I did not see such a flag when grep rho throughout the code. Maybe it has already been updated?

ED/src/init/ed_params.f90
xiangtaoxu: I intentionally did not change the existing sapwood biomass calculation. Sapwood is treated as bdead in my version, because I do not want cohorts to have a large carbon pool to burn off (sapwood accounts for ~70% of total wood volume for some tropical species) under negative carbon budget.

xiangtaoxu:
@mpaiao This should be the full pull request.

xiangtaoxu:
@tompowell9. I included your plant water stress scheme. You might want to test whether this version can reproduce your results as well.

xiangtaoxu:
@mpaiao Thanks for the review. I am actively using the hydro version for several projects, so I will create update PRs when necessary.

panditk17:
Hi @xiangtaoxu 
I am getting the same error as stated by Femeunier, about small tree flag. I am getting this error for PFT 5 (a grass PFT). I might have missed solutions above. 

image


Could you please let me know if I can solve this issue for grasses or small trees? Thanks.

xiangtaoxu:
@panditk17 
It seems your leaf water pool is overcharged before this timestep. Which version are you using? Is it after commit cf2e960?

In addition, I have only tested hydraulic parameterization for tropical PFTs. For other PFTs like PFT 5, I use trait correlations derived from tropical species to predict hydraulic parameters, which might be out of range and leading to errors..... Are you using the default parameters?

panditk17:
@xiangtaoxu 
Thank you for the response. I am using the latest version (I think that is after commit). I was using default parameters for PFT 5 for hydrodynamics. I was thinking the above error was mainly because smaller trees not matching up well with these hydro parameters. As you suggested, parameters based on simple correlations with tropical PFTs might not work well with grass PFTs.

xiangtaoxu:
@panditk17 OK, I will take a look at the codes later this week. How long have you run the model before it crashes? I have a quick comparison between PFT 5 and PFT 1, it seems that one big difference is hgt_min. Could you try set hgt_min for PFT 5 the same as PFT 1 (0.5) and run the simulation again?

PR #235, opened 2018-02-13
Merge recent ED2 changes into EDR

ashiklom:
## Description

Recent revisions to ED2 haven't made their way into the EDR code base. This PR applies them to EDR. The only change to ED2 is to modify `include.mk.gfortran` to use linker flags (e.g. `-lhdf5`) instead of direct file paths (e.g. `-L${HDF5_HOME}/libhdf5.so`) because these search through more possible places (e.g. `/usr/local/lib`, and anywhere else known to `ldconfig`), thereby making them a little more versatile.

## Collaborators

Shawn Serbin (@serbinsh), Michael Dietze (@mdietze) 

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue) 

## Checklist:

## Testing :


ED/build/make/include.mk.gfortran
```diff
@@ -48,9 +48,7 @@ USE_HDF5=1
 HDF5_HOME=/usr
 HDF5_INCS=-I${HDF5_HOME}/include
 HDF5C_INCS=-I${HDF5_HOME}/include
-HDF5_LIBS=${HDF5_HOME}/lib/libhdf5_fortran.a \
-        ${HDF5_HOME}/lib/libhdf5.a \
-        ${HDF5_HOME}/lib/libhdf5_hl.a -lz
+HDF5_LIBS=-lhdf5 -lhdf5_fortran -lhdf5_hl -lz
```
serbinsh: @ashiklom looks good.  My only wonder is whether this might cause issues on module systems where we just want it looking in the loaded model PATH and not across the entire system?  I suspect it wont be a problem since loading the module (e.g. module load hdf5/hdf5-version-something) updates $PATH and that is where it will search, correct?

serbinsh: I guess this is also only relevant if you use that mk file as your base for compiling, and perhaps not for machine-specific?

ashiklom: This is only the `gfortran` Makefile -- this doesn't affect the ones specific to `geo` or MODEX or anything else. Bot moreover, because `module` commands typically modify the `LD_LIBRARY_PATH` environment variable, this change should actually make it _more_ robust to that system. Before this change, this would _always_ look for the HDF5 libraries in `/usr/lib/libhdf5*`, so unless the `module` system symlinks to that directory, it would only ever load the default installation (if that was even where it was installed -- on some systems, like Ubuntu, the default install of HDF5 via `apt` gets placed in `/usr/x86_64-library/lib` or something like that).

serbinsh:
Perfect, thanks for the clarification.


On Feb 14, 2018, at 8:52 AM, Alexey Shiklomanov <notifications@github.com<mailto:notifications@github.com>> wrote:


@ashiklom commented on this pull request.

________________________________

In ED/build/make/include.mk.gfortran<https://github.com/EDmodel/ED2/pull/235#discussion_r168179367>:

> @@ -48,9 +48,7 @@ USE_HDF5=1
 HDF5_HOME=/usr
 HDF5_INCS=-I${HDF5_HOME}/include
 HDF5C_INCS=-I${HDF5_HOME}/include
-HDF5_LIBS=${HDF5_HOME}/lib/libhdf5_fortran.a \
-        ${HDF5_HOME}/lib/libhdf5.a \
-        ${HDF5_HOME}/lib/libhdf5_hl.a -lz
+HDF5_LIBS=-lhdf5 -lhdf5_fortran -lhdf5_hl -lz


This is only the gfortran Makefile -- this doesn't affect the ones specific to geo or MODEX or anything else. Bot moreover, because module commands typically modify the LD_LIBRARY_PATH environment variable, this change should actually make it more robust to that system. Before this change, this would always look for the HDF5 libraries in /usr/lib/libhdf5*, so unless the module system symlinks to that directory, it would only ever load the default installation (if that was even where it was installed -- on some systems, like Ubuntu, the default install of HDF5 via apt gets placed in /usr/x86_64-library/lib or something like that).

PR #240, opened 2018-03-15
Updates to fix build issue with EDR post addition of ED2 hydro

serbinsh:


serbinsh:
Working to fix issue #238.  There were new entries in objects.mk and rules.mk with the updates to ED2 to include hydro.  We needed to add those to the EDR version as well in EDR/build/make/

Testing build now

serbinsh:
@hamid4965

PR #253, opened 2018-07-31
Add MPI_Finalize call at the end of edmain

mrjohnston:
## Description

Added the following line in edmain.f90 for if (icall == 0) :
call MPI_Finalize(ierr)

## Motivation and Context

Without this MPI_Finalize call, the model does not finish cleanly on Harvard's Odyssey cluster. Though it produces output, it is considered 'failed' because 'mpirun has exited due to process rank 0 ... [node] exiting improperly'

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue) 
Will not change answers

## Checklist:

## Testing :

### DIFF
--- a/ED/src/driver/edmain.F90
+++ b/ED/src/driver/edmain.F90
@@ -290,6 +290,7 @@ program main
    end if
 #endif
    if (icall == 0) then
+      call MPI_Finalize(ierr)
       write(unit=*,fmt='(a)') ' ------ ED-2.2 execution ends ------'
    end if
    !---------------------------------------------------------------------------------------!
--- a/ED/src/driver/edmain.F90
+++ b/ED/src/driver/edmain.F90
@@ -285,13 +285,13 @@ program main
 
    !----- Finishes execution. -------------------------------------------------------------!
 #if defined(RAMS_MPI)
-   if (ipara == 1) then
+   select case (isingle)
+   case (0)
       call MPI_Finalize(ierr)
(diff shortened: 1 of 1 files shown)


mpaiao:
@mrjohnston I agree with @xiangtaoxu that the best option is to replace the block

```Fortran
#if defined(RAMS_MPI)
   if (ipara == 1) then
      call MPI_Finalize(ierr)
   end if
#endif
```

with

```Fortran
#if defined(RAMS_MPI)
   select case (isingle)
   case (0)
      call MPI_Finalize(ierr)
   end select
#endif
```

---
Lengthy explanation just because this may be a source of confusion for others.  It is important that MPI commands are always inside the #if defined(RAMS_MPI) blocks; these are pre-processor instructions, so the compiler will include the MPI commands only when
```Makefile
PAR_DEFS=-DRAMS_MPI
```
in include.mk.______ file. When 
```Makefile
PAR_DEFS=
```
the code will skip MPI altogether.  By the way, with the most recent versions of ED2, you don't need to set RAMS_MPI or use mpif90/mpicc or even have MPI installed, unless you are planning to run the coupled ED-BRAMS model.  MPI libraries are not needed for the -openmp option, the shared memory instructions work fine with regular gfortran/ifort.

xiangtaoxu:
@mrjohnston, can you make the changes suggested by Marcos in your branch? Then I can review and merge the change

PR #254, opened 2018-08-22
Add "observation time" output capability

mrjohnston:
## Description

These changes allow the user to output instantaneous files at specified times, rather than for the entire simulation (or none of it).   

## Collaborators

Xiangtao Xu

## Motivation and Context

This is useful when the user needs an "instantaneous" output at a particular point in time (perhaps associated with an observation), but does not wish to waste the space & time which would be required to produce instantaneous outputs for the entire simulation.  

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

This feature does not change any output values; it only changes what files may be outputted.

## Checklist:

There are two new options in ED2IN:
IOOUTPUT [0 if observation time output is off, 3 if it is on]
OBSTIME_DB [path to the observation time text file which specifies times to output]

This change will also necessitate an example OBSTIME_DB. I'm happy to do the updates if you let me know where.

## Testing :


xiangtaoxu:
@mrjohnston 
Great to see this coming! Some quick comments
1. Could you update the ED2IN templates in the ED folder and the EDTS folder by adding the new ED2IN option.
2. Could you also add a template for obstime_db file in the ED folder?
@mpaiao @rgknox Would you two have time to review the codes, given that you are familiar with the I/O part of ED? The biggest motivation for this new feature is to allow comparison with remote sensing data and maybe you will be interested as well.

mrjohnston:
Sure, thanks. Edited!

mpaiao:
I checked the code and most of it seems fine with me.  Before I approve the merge, I have two minor suggestions, both in ed_opspec.F90.

1.  When both fast and observation output files are on, shouldn't this cause a fatal_error instead?  In case you think a warning is fine, I suggest replacing the write statement (lines 1303-1305) with a call to sub-routine warning (available in src/utils/fatal_error.f90)

2. Shouldn't we include a check that causes the model to crash in case IOOUTPUT = 3 and OBSTIME_DB is empty?

mrjohnston:
Thanks a lot, Marcos. Regarding your suggestions:

1. I prefer a warning because IOOUTPUT is nested in IFOUTPUT. If they're both on, the user will indeed get the output that s/he has requested in OBSTIME_DB, but there will also be additional, potentially extraneous outputs. So: I don't see a reason to crash. I'm now calling the 'warning' subroutine rather than using a write statement, thanks for the suggestion.

2. Agreed. This check occurs within the read_obstime subroutine (init/ed_init.F90, lines 1165-1171).

PR #224, opened 2018-01-21
Update to travis #219

robkooper:
Use Linux (trusty container) and OSX
Install packages Travis way

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue) 
- [ ] New feature (non-breaking change which adds functionality)
- [X] Other (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.

## Testing :

- [ ] All new and existing tests passed.

Issue #219, opened 2018-01-17
Test up Travis automated testing on PRs


bpbond:
Steps:
1. Compiles successfully
2. ...with multiple compilers and platforms
3. ...with and without multithreading and MPI
4. ...runs simple test case
5. ...strict (no warnings)

Names on whiteboard about this: me, Marcos, Rob, and "XX"  cc @mdietze

bpbond:
Tested existing `.travis.yml` file and it compiles ED2 correctly.

robkooper:
Now remembered, this been around for a while, see https://travis-ci.org/EDmodel/ED2/

PR #281, opened 2019-09-18
Fix Travis CI build

ashiklom:
## Description

- Use `apt` add-on instead of manual `sudo apt-get ...` commands
- Update HDF5 paths to reflect `apt` install locations

With these changes, I am able to at least get a [successful build](https://travis-ci.org/ashiklom/ED2/builds/586679635) on Travis. It would also be nice to do some simple model runs, but that's much more involved (mostly because you need to get inputs from somewhere.)

## Collaborators

- Author: @ashiklom
- Supersedes PR #224 by @robkooper

## Motivation and Context

Current build was breaking because of invalid HDF5 paths in `include.mk.travisci`.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue) #219

## Checklist:

## Testing :

- [X] All new and existing tests passed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,20 @@
 language: c
-sudo: required
+dist: xenial
+
+addons:
+  apt:
+    packages:
+      - gfortran
+      - libopenmpi-dev
+      - libhdf5-openmpi-dev
 
 before_install:
-    - sudo apt-get clean && sudo apt-get update
-    - sudo apt-get install gcc gfortran hdf5-tools libopenmpi-dev libhdf5-openmpi-dev 
     - which gfortran
+    - gfortran --version
     - which gcc
-    - which h5dump
+    - gcc --version
 
 script:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,14 +1,20 @@
...
-    - sudo apt-get install gcc gfortran hdf5-tools libopenmpi-dev libhdf5-openmpi-dev 
     - which gfortran
+    - gfortran --version
     - which gcc
-    - which h5dump
+    - gcc --version
 
 script:
-    - pwd
+    - set -e
```
robkooper: good catch, otherwise travis will compile with an error and still give thumbs up.

ashiklom: I [learn from the best](https://github.com/EDmodel/ED2/pull/224/files#diff-354f30a63fb0907d4ad57269548329e3R27).

PR #282, opened 2019-09-19
New Forestry Routine Features

crollinson:
## Description

Two main changes to the forestry routines:
1. New IANTH setting where 2 means f_vbh becomes the fraction area harvested rather than trying to achieve biomass harvest targets.  (It's hard to set a biomass harvest target when you're not sure what the long-term productivity will be -- much easier to think about areas harvested.)
2. Add code to allow harvest to occur below DBH threshold to better facilitate selective harvesting regimes like understory thins or timber improvement thins.

This also begins to reconcile differences in the implementation of the plantation code, following what @mpaiao had done.  "Secondary" forest is now references plantations in most parts of the LU code, although I have not reviewed/modified plantation behavior while "primary" forest is just natural forest.

Note: There are more changes coming eventually, including updates to the plantation schemes and potential to have multiple management schemes operating at once.  Plantation updates probably within the next 6 months, multiple management is a long ways off.

## Collaborators

Most of what I did is operationally or conceptually consistent with what Marcos Longo (@mpaiao) did a long time ago that never got pulled into the mainline.

## Motivation and Context

These changes are to facilitate work of an NSF-funded project focused on interactions between forest management & climate change.  These changes _should_ help make ED behavior more compatible with how US forest managers think about harvesting routines by adding more control.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)
- [X] Other (fix or feature that would cause existing functionality to change)
_*NOTE:* Headers for land use drivers will need to be updated to add 2 lines for the harvest rates below the DBH threshold.   [See example](https://github.com/MortonArb-ForestEcology/MANDIFORE_modeling/blob/master/WCr_Test_UnderstoryThin/HarvestBelow/US-Wrc_HarvestBelow_lat45.5lon-90.0.lu).  Based on what I could find, LU driver templates are not currently in any of the ED folders nor test cases, but I can provide one if necessary._

## Checklist:

- [X] I have added tests to cover my changes.

## Testing :

None of the routine ED tests include land use change or forest harvesting.  However, here are two files showing change in AGB and stem density in a short test of different selective harvests that I ran.

[HarvestDifferences_AGB.pdf](https://github.com/EDmodel/ED2/files/3632102/HarvestDifferences_AGB.pdf)
[HarvestDifferences_Density.pdf](https://github.com/EDmodel/ED2/files/3632103/HarvestDifferences_Density.pdf)

Code for above tests at Willow Creek with ED2INs & LU drivers associated with above can be found here: https://github.com/MortonArb-ForestEcology/MANDIFORE_modeling/tree/master/WCr_Test_UnderstoryThin


mpaiao:
@crollinson Just to give feedback on this merge.  We discussed this merge at the ED2 workshop, and we will merge once the ED-2.2 merge is incorporated into the mainline.

mdietze:
Just wanted to follow up on this PR. @crollinson there's now a lot of conflicts after we pulled in @mpaiao ED2.2 release code. Wanted to check in to see when you think you'll get to those.

mpaiao:
I started to look into and work on this merge, but I will be quite busy in the next two weeks. I don't anticipate this being a difficult merge — the conflicts look big, but it's more that the code was restructured (especially ed_params.f90) and the changes are in different locations.  It will require some manual merge, though, and I will likely need help from @crollinson to make we don't miss anything in the process.

crollinson:
I'll try to do this Monday next week -- this week is not going to be a good week.

ED/src/dynamics/mortality.f90
```diff
@@ -273,16 +279,26 @@ real function survivorship(new_lu,dist_path,mindbh_harvest,cpatch,ico)
...
          ! different pathways for conventional and reduced-impact logging could be         !
          ! applied.                                                                        !
          !---------------------------------------------------------------------------------!
          if (cpatch%dbh(ico) >= mindbh_harvest(ipft)) then
-            survivorship = 0.0
+            !if (ianth_disturb == 2) then
+            	survivorship = 1 - harvprob_g(ipft)
+            ! else 
+            !	survivorship = 0.0
+            ! end if
```
mpaiao: @crollinson I am going through the code merging but this part is a bit confusing to me., and I think this approach here may be saving too many trees.  In my original implementation, harvprob is used to define the disturbance rate.  In general, survivorship should reflect the survivorship in the immediate vicinity of disturbed area (in this case, where the tree was logged), which is why survivorship was set to zero.  For the time being I kept your approach for the area-based option (ianth_disturb=2), but I am not sure how to reconcile this part with the original survivorship definition, especially with understory thinning.

mpaiao: The only idea I had so far is to create a new dist_type for thinned canopy.  The upside is that it would be consistent with the ED idea that disturbances change the microenvironment (in the case of thinning, more in the water part than in the light).  The one potential downside is that we would have more dist_types (tree felling, skid trails and roads, and artificially thinned canopy).

I don't see a problem to add a new dist_type if it is justified, but before I go ahead, I'd like to throw the idea and see what people think, or if anyone has other ideas.

Issue #285, opened 2019-10-02
Soil evaporation calculation


xiangtaoxu:
I noticed that current mainline uses saturated specific humidity (ground_ssh) instead of true specific humidity (ground_shv = ground_ssh * soil_moisture factor) when calculating soil evaporation
https://github.com/EDmodel/ED2/blob/7a39cd9de80d5551acd0753ee94470ca6facb933/ED/src/dynamics/rk4_derivs.F90#L1255-L1257

This can lead to unrealistically high soil evaporation during tropical dry seasons, creating extra plant water stress. I also checked earlier versions (rv84), which uses ground_shv instead of ground_ssh.

Just want to confirm this should be ground_shv. @mpaiao

mpaiao:
As far as I remember, this should indeed be ground_shv, not ground_ssh.  Not sure why this has changed...  Thanks for checking this.

xiangtaoxu:
I have just created a quick PR for this. I also urge people who run the master version in hot and dry ecosystems to check their evaporation and soil moisture values during dry months.

PR #286, opened 2019-10-03
[Bug Fix] use ground_shv to calculate ground evaporation

xiangtaoxu:
Calculate ground evaporation using actual ground surface specific humidity. Previously, it was calculated using saturated specific humidity (ground_ssh), which can overestimate evaporation during hot dry period, and potentially create numerical instability.

## Description

Only one change in rk4_derivs.f90. Other uses of ground_ssh in the subroutine are valid (for dew formation calculations)

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue) 
#285 

## Checklist:

## Testing :


mpaiao:
I will double check, but I am pretty sure this fix was independently incorporated in the PR #289.

mpaiao:
I checked and this fix is already incorporated in the mainline.  @xiangtaoxu is it fine if we close it?

Issue #288, opened 2019-11-05
The ED-2.2 Merge


mpaiao:
Opening this issue to discuss the ED-2.2 merge to the mainline, following the ED2 meeting in Nov 2019.

mrjohnston:
**Compilation:** successful (platform: Cannon; compiler option ("kind"): E)

**Running:** terminates before any output is written with the following error:
 [+] Ed_Init_Atm...
OLD NPLANT:   1.0001395E+00
NEW NPLANT:   5.3516275E-01
OLD SIZE:     2.8055266E-01
NEW SIZE:     1.5028669E-01
::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::

--------------------------------------------------------------
                     !!! FATAL ERROR !!!
--------------------------------------------------------------
    ---> File:        fuse_fiss_utils.f90
    ---> Subroutine:  split_cohorts
    ---> Reason:      Cohort splitting didn't conserve plants!!!
--------------------------------------------------------------

ashiklom:
@mrjohnston Would you be able to share your inputs and configuration for these runs with me? If so, I can add them to the new Continuous Integration test suite.

mrjohnston:
@ashiklom That would be lovely, thank you! One quick caveat is that I wouldn't necessarily expect results to be great, since the phenology is specified incorrectly, the initial condition is simplified, etc. 
That said, this configuration with these inputs completes successfully using ED2.1 (but not using ED2.2 as of November -- as you can see above).

[Tonzi_InputsConfig.zip](https://github.com/EDmodel/ED2/files/4032015/Tonzi_InputsConfig.zip)

mrjohnston:
@ashiklom That would be lovely, thank you! One quick caveat is that I wouldn't necessarily expect results to be great, since the phenology is specified incorrectly, the initial condition is simplified, parameters are not fully tuned, etc. 

That said, this configuration with these inputs completes successfully using ED2.1 (but not using ED2.2 as of November -- as you can see above).

[SemiArid_InputsConfig.zip](https://github.com/EDmodel/ED2/files/4032050/SemiArid_InputsConfig.zip)

mpaiao:
@mrjohnston thanks for sending the crash case.  I may try to incorporate to my local tests too.

[...]

PR #289, opened 2019-11-18
The ED-2.2 merge pull request

mpaiao:
## Description

This is the pull request of ED-2.2, as discussed in the ED2 workshop in Boston in November 2019.  Below is a summary of changes:

- Energy, water, carbon and CO2 are now checked for conservation with stricter tolerance and at the dynamic time steps.
- Disturbance is now calculated in a way to make total disturbed area consistent with disturbance rate.
- Growth rates that account for the need to grow living tissues in the upcoming month.  This resolves most of the issues of setting a reasonable qsw, pointed out by @yjkim1028 ages ago.
- Internal water is now completely separated from oven-dry biomass in leaves and sapwood, to properly account for energy exchange when plant hydraulics is dynamic.
- ed_params.f90 reorganised to separate parameters that can and cannot be initialised through xml.
- New patch and cohort fusion scheme that allows ED-2 to be initialised with thousands of patches. 
- Several new optional features such as Jmax based light limitation, TPU limitation on photosynthesis, new allometry based on DBH^2*H, height-dependent reproduction, five-pool decomposition scheme.
- Several house-keeping changes in the code, like removal of all interface blocks, which are now replaced by modules, enforcement of "_use, only_" usage in 99% of routines (ed_xml_config.f90 is pretty much the one big exception).

## Collaborators

Marcos Longo (@mpaiao), with _lots_ of help from Xiangtao Xu (@xiangtaoxu).

## Motivation and Context

This is a list of changes over the past few years, in order to enforce conservation of most properties in the ED-2 model, and expand the model's ability to simulate complex disturbance regimes caused by anthropogenic activities.  Most of the changes are described in the following publication:

**Longo M, Knox RG, Medvigy DM, Levine NM, Dietze MC, Kim Y, Swann ALS, Zhang K, Rollinson CR, Bras RL, Wofsy SC, Moorcroft PR. 2019.** The biophysics, ecology, and biogeochemistry of functionally diverse, vertically and horizontally heterogeneous ecosystems: the Ecosystem Demography model, version 2.2 – part 1: Model description. _Geosci. Model Dev._ **12**: 4309–4346, [10.5194/gmd-12-4309-2019](https://dx.doi.org/10.5194/gmd-12-4309-2019).

## Types of changes

## Checklist:

As discussed in the ED2 workshop, we will need help from others testing this pull request. Feedback will be much appreciated.

Most of the documentation is updated in the @mpaiao  [wiki](https://github.com/mpaiao/ED2/wiki) and I already updated part of the wiki here. Once the changes are accepted, I will incorporate the relevant changes in the wiki.

## Testing :


mpaiao:
@mdietze The subroutine is init_derived_params_after_xml, it's the last subroutine in EDBRAMS/ED/src/init/ed_params.f90

ED/src/init/ed_params.f90
```diff
@@ -4570,1076 +7248,1598 @@ end subroutine init_hydro_coms
...
+   ! the calculation from earlier versions of ED-2.2.  We now separate oven-dry biomass    !
+   ! and internal water for leaves and sapwood, as the latter is dynamic.  For now heart-  !
+   ! wood and bark are assumed to have constant internal water content, so we incorporate  !
+   ! the specific heat of internal water to the bulk specific heat.                        !
+   !     In addition, we ignore the wood-water bonding heat capacity for sapwood.  This is !
+   !  a relative small effect (2-5% of heat capacity), and it only changes when water      !
+   !  content is very low.  Keeping this term may add some non-linearities and would       !
+   ! require dynamic updates --- not a big deal, and we may add this effect in future      !
+   ! versions.                                                                             !
+   !
```
mdietze: @serbinsh @ashiklom  starting here, could you review that the changes @mpaiao made reflect things that should be calculated from other ED optics variables and should NOT be set by the XML. We should make sure this is true for both the default code and RTM stand alone.

ashiklom: I'll try to take a look at this later this week.

ED/src/init/ed_params.f90
```diff
@@ -2314,1771 +5945,936 @@ end subroutine init_pft_mort_params
...
+   zetac_uhi13 = 1.0 / cbrt(-zetac_uh)
 
-   !------------------------- Liana leaf Biomass (Putz, 1983) -----------------------------!
-   b1Bl_small(17) = b1Bl_small(2)
-   b2Bl_small(17) = b2Bl_small(2)
-   b1Bl_large(17) = 0.0856
-   b2Bl_large(17) = 2.0
-   !---------------------------------------------------------------------------------------!
 
+   !----- Parameters for the z0m:z0h ratio, following Zeng and Dickin
```
ashiklom: This is a good place to define what `vis` and `nir` mean in terms of wavelength ranges. Otherwise, it's difficult to say how realistic these values actually are.

In my experience with the remote sensing literature, "visible" is about 400-700/750nm, "near-infrared" is from the end of visible to about 1400nm, 1400-2500 is "shortwave-infrared" (SWIR; which ED2 doesn't have), everything from there to like 5000nm is "mid-infrared" (MIR), and from there on is thermal.

@serbinsh @mrjohnston -- thoughts?

mrjohnston: Yep, pyrgeometers typically measure from 4.5 or 5 micrometers.

mdietze:
So this PR has been open a while. I'm going to propose that any further discussion of these changes be limited to this week at AGU and that there's no objections that we pull this at the end of the week.

mpaiao:
Updates to the pull request.  I made changes to the code to resolve some issues with energy conservation in mid-latitudes or when flooding occurs.  Details below:

1. Bug fixes in the energy and water conservation, likely to be triggered with cold-deciduous phenology and/or bedrock.

- Phenology.  Leaf surface water was being set to zero when cohorts shed all the leaves (phenology_status=-2).  The cold-deciduous phenology is a sudden transition, and if it rained/snowed the day before the leaf-dropping day, the intercepted water would go unaccounted and cause water/energy conservation issues. I fixed this by accounting for this loss of water.
- Temporary surface water.  The leak was happening when the TSW layer was terminated.  Radiation is calculated every dtlsm, but the thermodynamics has sub-steps.  When the TSW layer is terminated in the middle of the radiation, the pre-calculated radiation of the terminated layers was going nowhere, causing the leak.  I fixed this by defining "commited" radiation terms, and sending them to either the topmost remaining TSW, or to the top soil layer if TSW is completely terminated.

2. Additional bug fixes and new features implemented while I was trying to fix the leak.
- Canopy air space density is now updated every sub-step.  This turned out not to be the bug, but updating every time step ensures ideal gas law and improves accuracy of enthalpy, water, and CO2 conservation in the canopy air space.
- Check for radiation conservation.  The original energy check only accounts for the net absorbed radiation, but it did not check conservation of radiation in the two-stream model.  The new radiation check is a new, separate test that ensures that Incident - Outgoing = Absorbed - Emitted, for every time step.
- The net absorbed shortwave radiation was not being correctly set or reset when the number of TSW layers changed, and outgoing longwave radiation was not updated when there were no active cohorts (bare ground simulations, or cases in which branch thermodynamics is turned off and all cohorts have phenology_status = -2).
- The code now excludes non-resolvable cohorts from the budget check.  We do not solve them prognostically in the thermodynamics, so they should not be included as storage for water and energy.  In the case of deciduous plants, they may enter and exit the group of active cohorts.  To account for this, I included new budget variables (ebudget_pheneffect/wbudget_pheneffect) to capture the transition between resolvable/non-resolvable cohorts in the energy and water cycles.
- Seedling_mortality when repro_scheme = 0.  When reproduction is disabled, we assign seedling_mortality = 1.0 for all PFTs.  This was done in reproduction.f90, but it created inconsistency because seedling_mortality is also used in structural_growth.f90, which is called before the first reproduction call.  I moved the check to ed_params.f90, in subroutine init_derived_params_after_xml.
- I moved some sub-routines from rk4_driver.f90 to rka4_copy_patch.f90.  They are sub-routines that copy variables, and are used by other integrators.  This has no impact on the simulations.

ashiklom:
I don't have the time or ability to review all the changes in this PR, but based on the fact that this passes my basic CI tests, I think we can merge this and start testing it more aggressively. We should add additional CI tests (different sites, install configurations, and ED2IN settings, preferably the factorial combination) and resolve bugs in smaller PRs as they come up.

ashiklom:
Update: I added @mrjohnston 's test from #288 (see PR #295) and have yet to get that to run all the way through. That configuration is currently failing in the `fuse_2_cohorts` subroutine, at line 2522 (fusing plant hydrodynamics characteristics).

https://github.com/EDmodel/ED2/blob/228b4900eb6e4ca677ad02b2af25c8db944aee54/ED/src/utils/fuse_fiss_utils.f90#L2521-L2524

In `lldb` (the MacOS debugger), the error looks like this:

```
 [+] Update_met_drivers...
 [+] Ed_Init_Atm...
Process 91375 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_ARITHMETIC (code=EXC_I386_SSEEXTERR, subcode=0x1132)
    frame #0: 0x0000000100abcdc8 ed_2.2-dbg`__fuse_fiss_utils_MOD_fuse_2_cohorts at fuse_fiss_utils.f90:2523
   2520	                                  + cpatch%wflux_wl     (donc) * dnplant
   2521	      do isl = 1,nzg
   2522	         cpatch%wflux_gw_layer(isl,recc) = cpatch%wflux_gw_layer(isl,recc) * rnplant       &
-> 2523	                                         + cpatch%wflux_gw_layer(isl,donc) * dnplant
   2524	      enddo
   2525	      !------------------------------------------------------------------------------------!
   2526
```

Doing some debug-by-print, it looks like the issue is when the flux is, for some reason, greater than zero, but by a miniscule (1e-39) amount. I'm guessing what's going on is that numbers that small throw a floating point arithmetic error:

```
Backtrace for this error:
 rnplant:   0.500000000
 dnplant:   0.500000000
 isl:            1
 recc:            2
 wflux(recc):    0.00000000
 wflux(donc):    0.00000000
 isl:            2
 recc:            2
 wflux(recc):    0.00000000
 wflux(donc):    0.00000000
 isl:            3
 recc:            2
 wflux(recc):    0.00000000
 wflux(donc):    0.00000000
 isl:            4
 recc:            2
 wflux(recc):    3.76525534E-39
 wflux(donc):    3.76525534E-39
#0  0x1546e3f3d
#1  0x1546e334d
#2  0x7fff72255b5c
...
```

mpaiao:
@ashiklom I solved this issue in the version I am working on for the test suite.  Perhaps I just forgot to push it to GitHub.  This variable is missing in subroutine copy_sitetype (ed_state_vars.F90), so it would become uninitialized when a patch was copied.  Let me update the pull request.

mpaiao:
I pushed my latest commits, which should fix the issue with wflux_gw_layer.  There were a few other minor bugs that I spotted and fixed when I ran the test suite.

ashiklom:
Ah, OK. I'll pull in these changes and see how they do.

ashiklom:
OK, so with those changes, the run does proceed for 1 month, but at the end of that month, it throws another arithmetic exception, this time in the plant hydrodynamics:

```
Thread 1 "ed2" received signal SIGFPE, Arithmetic exception.
0x0000000000be6560 in plant_hydro::rwc2tw (leaf_rwc=1, wood_rwc=1, bleaf=4.48415509e-44, bsapwooda=4.48415509e-44,
    bsapwoodb=4.48415509e-44, bdeada=4.48415509e-44, bdeadb=4.48415509e-44, broot=4.48415509e-44, dbh=0.502384722,
    ipft=5, leaf_water_int=0, wood_water_int=1.77273e-37) at plant_hydro.f90:1157
1157	      leaf_water_int    =   leaf_rwc * leaf_water_sat(ipft) * bleaf * C2B
```

Looks like a similar issue here: `bleaf` has biomass pool values on the order of 1e-44, which causes floating point arithmetic errors.

mpaiao:
@ashiklom @mrjohnston I merged the pull requests #294 and #295, and found a few other bugs in ed_read_ed10_20_history.f90.

The test suite is running, but I noticed that the model is stuck on October 3, 2001 (for over 30 minutes).  I assume that the time step is shrinking and the test will eventually crash.  Was this happening in your version too when you committed #295, or did it run until the end? I need to run now, and I will try to take a look at it later or next week, but if you have any ideas of the possible causes, please let me know or give a try with the latest version. Thanks!

ashiklom:
@mpaiao Thanks for the quick fixes! And looks like the run succeeded! I think the live CI log just gets stuck sometimes, especially for long logs.

The Tonzi runs also worked for me on my Mac desktop (`-p macos_homebrew`), which is encouraging.

https://github.com/EDmodel/ED2/pull/289/checks?check_run_id=384032765

mpaiao:
I ran the test suite, medium length — about 125 years for individual sites and 10 years for budget. The automated report is available [here](https://github.com/EDmodel/ED2/files/4082818/653299e-mlo-5b922ea-v1_medium.pdf).  I have pushed additional commits since 5b922ea, but these were all to fix the test suite scripts, not the ED-2 code. 

Most simulations successfully finished using the pull request version, whereas the mainline had more crashes.  The only exception for the pull request was Manaus Km 34, which uses hybrid time step.  The Manaus Km 34 simulation also crashed for the mainline, although it did because of numerical instability.  The hybrid subroutines are a bit of a black box to me, and I had to change it because of the separation of internal water and oven-dry biomass for leaf and wood internal energies (consequence of plant hydrodynamics). In case anyone is more familiar with the hybrid scheme and is willing to take a look, it would be very helpful.

Budget closure plots don't look very different between mainline and pull request.  This is expected, because most of the changes in the budget closure were in the vegetation dynamics time steps (phenology, structural growth, etc), which happen infrequently and were causing spikes in the residual.  Also, to be comparable, I set both the mainline and the PR version to use the same tolerance for budget (about 1e-4).  I tried with 1e-5 too, and there are some cases associated with long-term dynamics that are causing the model to crash in the pull request.  I hope to address these additional crashes in future pull requests.  That said, nearly all the mainline simulations crashed with 1e-5, so the pull request has actual improvements in this regard.  Also note that carbon budget is a new feature in the pull request, so there is no comparison now, but the code will be able to compare versions in the future.

As I mentioned before, there were changes in the code, so we should not expect the results to be identical between mainline and test suite. Although most of the changes were in thermodynamics, they affect pretty much everything else. In any case, there were some noteworthy differences that may need additional investigation:
1.  Harvard forest seems to be losing conifers now.  It may be some mismatch in settings in ed_params.f90.
2. LAI in the tropics is generally higher in the pull request.  The total values make more sense, but a lot of them are coming from small trees.

I will be offline for a few days.

PR #290, opened 2019-12-27
Continuous integration with GitHub Actions

ashiklom:
- Build ED2

- Run a 1-year simulation at UMBS from bare ground, and check to make sure it completed successfully

I've left instructions for adding additional tests in the `README.md` file. I figure we can add more tests in downstream PRs.

Looks like this repo's CI won't run until this PR has been merged. Here is what the results look like from my fork: https://github.com/ashiklom/ED2/actions


mdietze:
Are there additional thing I need to do as an admin to turn on these CI checks as a requirement for pulling pull requests (currently there's just Travis) or will that happen automatically when this is pulled?

ashiklom:
Not sure. I'm fairly sure that the checks will always run once this is merged, so there's always the possibility of manually checking that everything succeeded. Ensuring that branches can't get merged before checks pass involves branch protection rules: "Settings" -> "Branches" -> "Branch protection rules" -> "New rule" -> "Require status checks before merging". You may need to check an additional box in the list of checks that will display.

PR #295, opened 2020-01-09
Add Tonzi semi-arid tests

ashiklom:
This PR adds a test at @mrjohnston 's site (See https://github.com/EDmodel/ED2/issues/288#issuecomment-571725827). It's a nice complement to the current UMBS test because, unlike that test, it includes vegetation initial conditions and a parameter XML file.

This PR builds on #294, so merge that one first.


ashiklom:
@mrjohnston Looking more closely, the code that ED2.2 is calling when it throws this error is:

https://github.com/EDmodel/ED2/blob/1c0318d4d625b0fd8f29828a6f07181b4c42e9d0/ED/src/init/ed_params.f90#L7932-L7933

With your default values, this means that:

```
d2 = 1 / 0.0001 = 10000
d1 = (2 / 0.0001) ^ 10000 = 20000 ^ 10000 => Huge!
```

In R, 20000 ^ 10000 produces `Inf`, so I'm guessing that's where the Fortran floating point arithmetic breaks down as well.

mrjohnston:
@ashiklom Yep -- that seems likely. Those equations are absent from the older version, in which there's no issue:

dbh2bd = b1Bs_small / C2B * dbh ** b2Bs_small 
= 0.0001 / 2 * dbh ^ 0.0001 
=~ 0.0001 / 2

Thanks for honing in!

PR #297, opened 2020-01-31
Fix metadata for Mortality rates

ashiklom:
Small metadata mismatch I noticed. Mortality rates are stored as matrices -- mortality class x cohort, but metadata says they are only by cohort.


ashiklom:
Closing this to re-direct it the master branch.

Issue #300, opened 2020-02-04
XML-prescribed events violate C budget checks


ashiklom:
A prescribed disturbance event (via `events.xml`) like this:

```xml
<?xml version="1.0"?>
<!DOCTYPE config SYSTEM "ed_event.dtd">
<eventlist>
  <event>
    <year> 2004 </year>
    <doy> 187 </doy>
    <harvest>
      <agb_frac> 0.1 </agb_frac>
      <bgb_frac> 0.1 </bgb_frac>
      <fol_frac> 0.1 </fol_frac>
      <stor_frac> 0.1 </stor_frac>
    </harvest>
  </event>
</eventlist>
```

...breaks the C budget in the following day.

```
  Carbon budget
 .................................................. 
 TOLERANCE         :   4.9705604E-06
 RESIDUAL          :  -1.6472839E-04
 INITIAL_STORAGE   :   4.0054037E-03
 FINAL_STORAGE     :   3.8411987E-03
 DELTA_STORAGE     :  -1.6420498E-04
 DENSITY_EFFECT    :   5.2919859E-07
 SEEDRAIN          :   0.0000000E+00
 CANDEPTH_EFFECT   :   0.0000000E+00
 LOSS2YIELD        :   0.0000000E+00
 LOSS2ATM          :   5.7805196E-09
 COMMITTED         :   2.4879503E-06
```

Marcos proposed the following fix:

> With the new C budget check, if the carbon is removed from the patch, then this loss should be documented. We could either add it to cbudget_loss2yield(ipa), or create a new cbudget variable if you think it's really necessary to distinguish them.  Using cbudget_loss2yield(ipa), you could add the following to the routine:

```fortran
csite%cbudget_loss2yield(ipa) = csite%cbudget_loss2yield(ipa) + biomass_removed_from_patch * frqsumi
```

ashiklom:
This has been addressed for the Carbon budget for harvest events in #301. Because the events code predates the budget checks, other events will likely fail in similar ways (e.g. irrigation will break the water balance).

Here's a list of events for tracking purposes. We can try to check these off as we test them.

- [X] Harvest -- Implemented in #301
- [ ] Fertilize
- [ ] Till
- [ ] Planting
- [ ] Irrigation
- [ ] Pesticide

The Events documentation is [here in the ED2 wiki](https://github.com/EDmodel/ED2/wiki/Event-files).

PR #301, opened 2020-02-04
Apply changes from harvest to C budget checks

ashiklom:
When harvest events (set via `event.xml`) happen, add the resulting change in C budgets to the `csite%cbudget_loss2yield(ipa)` term. This at least partially addresses #300 (it only deals with harvests -- not sure how it'll handle other events).

This also adds a new instance of the Tonzi test that includes a partial harvest event partway through.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,3 +52,20 @@ jobs:
       run: |
         chmod +x "$GITHUB_WORKSPACE"/ed2-binary/ed_2.2-dbg
         ./run-test.sh tonzi "$GITHUB_WORKSPACE/ed2-binary/ed_2.2-dbg"
+
+  test-tonzi-harvest:
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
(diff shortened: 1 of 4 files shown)


ashiklom:
Here are the current CI results: https://github.com/ashiklom/ed2/actions/runs/34657659.

Not sure why they aren't running for this pull request. We should check that they are set up correctly.

ashiklom:
I think the checks are a relatively new feature (and a good one at that!).

I'm going to punt on fixing the rest of the events -- I wanted to get this fixed quickly for the person taking over from my disturbance modeling work with Ben and Chris, but I don't have the time to do the others, at least right now. I'll make a note in the issue that the other budgets are currently unresolved.

mpaiao:
@mdietze I don't think all the cbudget variables existed, and even if some of them did, they were not being used to verify carbon conservation. I added several of them and wrote the routines that check for C conservation as part of the ED-2.2 papers. 

Because I never used the events.xml file, I was not aware that they were applying changes to existing patches. Yes, irrigation must be accounted in both water and energy, there are checks for them too.

PR #305, opened 2020-03-31
Updates in heterotrophic respiration (DECOMP_SCHEME=5 only), and bug fixes in plant hydrodynamics and patch fusion

mpaiao:
Changes in heterotrophic respiration functions for DECOMP_SCHEME=5, to improve seasonal variation in the tropics.

Bug fixes in plant hydrodynamics (to prevent RK4 crashes) and patch fusion (to prevent crash due to canopy air space properties not conserving CO2).

## Description

1. Updates in the environmental (temperature and soil moisture) controls on heterotrophic respiration, when using the CENTURY-based (5-pool) approach (DECOMP_SCHEME=5).  This now uses a formulation based on [Koven et al. (2013)](https://dx.doi.org/10.5194/bg-10-7109-2013), namely a Q<sub>10</sub>-approach for scaling temperature, and a down regulation moisture and oxygen limitation that are based on soil matric potential instead of soil moisture.  These changes do not interfere with other DECOMP_SCHEME approaches. 

2. Added more decomposition-related variables to ED2 output, specifically the monthly average inputs to metabolic (fgc_in, fsc_in) and structural litter (stgc_in, stsc_in).

3. As I was implementing 1 and 2, I came across a few issues in the model, so I fixed some bugs.

- Soil conductance could be large when soils are dry, because of a hard minimum conductivity.  I now disabled the minimum by default, although it can still be used for hydrodynamics off.  Also, in plant_hydro.f90, soil-root conductance was effectively zero when soil moisture was below soilcp (air-dry residual moisture). Because hydrodynamics is not updated in RK4, soil moisture could drop below soilcp causing crashes. I fixed this by preventing conductance to occur when soil moisture approaches soilcp.  I wonder if conductance should go to zero at the permanent wilting point instead (question for @xiangtaoxu).   
- Similarly, the minimum leaf water potential had to be capped at slightly higher values than the hard minimum, to prevent potential of previous time step to become positive (which caused plant_hydro.f90 to issue an error).
- In patch_fusion, I rewrote the order of the operations when fusing the canopy air space properties, to make sure that the total content of CAS properties was preserved.  This was mostly implemented, but some minor non-linearities in the total mass were causing deviations in the conservation of CO2 to reach 10<sup>-4</sup> in some cases, which was enough to issue an error. 

## Collaborators

@mpaiao and inputs from @xiangtaoxu.

## Motivation and Context

1. Heterotrophic respiration changes: improve seasonality and magnitude of ecosystem respiration in the tropics, and provide additional information on the necromass carbon cycle.
2. Plant hydrodynamics: improve stability of the model.
3. Patch fusion: conserve canopy-air-space CO2. 

## Types of changes

## Checklist:

- [x] I have updated the documentation accordingly.

I did not run the full test suite because changes were specific to DECOMP_SCHEME=5 (considered beta for now), or were bug fixes.  Simulations that were crashing before the bug fixes are now running fine.

## Testing :

- [x] All new and existing tests passed.


femeunier:
Not sure it is the best place to mention this, but I was trying the pull request for my own simulations as I was experiencing issues with the hydro version in very dry conditions. 

While the merges solved most of the troubles I had (especially those related to soil conductance in dry conditions), I noticed that I could no more use the restart option: the model is crashing at the very beginning of the second day of simulation due to Sub-daily budget failures. I attach my [ED2IN](https://github.com/EDmodel/ED2/files/4456477/ED2IN_history.txt), the [crash log](https://github.com/EDmodel/ED2/files/4456476/BCI_out.32849855.txt) and [an example of restart file](https://github.com/EDmodel/ED2/files/4456475/BCI-S-2015-02-01-000000-g01.h5.txt) but I suspect it might be the case for most of the restart runs.

mpaiao:
@femeunier It looks like all patches are crashing at the same time, and from the output, the budget check is pointing to carbon not being conserved. It's hard to say exactly what is happening because all patches are dumping the message to the standard output at the same time.  My suspicion is that some variable is not being properly read from the restart file. It is likely to be a variable that is not used until the daily time step, hence the crash in the second day.  To help figuring out what is going on, would you mind doing the following?

1.  Temporarily, go to ED/src/utils/budget_utils.f90, look for the following _if_ block:
```
         if (.not. budget_fine) then
            write (unit=*,fmt='(a)') '|--------------------------------------------------|'
            write (unit=*,fmt='(a)') '|       !!!    Sub-daily budget failed    !!!      |'
            write (unit=*,fmt='(a)') '|--------------------------------------------------|'
            write (unit=*,fmt='(a,i4.4,2(1x,i2.2),1x,f6.0)') ' TIME           : ',         &
               current_time%year,current_time%month,current_time%date ,current_time%time
            write (unit=*,fmt=fmti ) ' IPA               : ',ipa
            write (unit=*,fmt=fmti ) ' DIST_TYPE         : ',csite%dist_type(ipa)
 ```
and replace unit=* with unit=20+ipa for all the lines in the _if (.not. budget_fine)_ block.  This will write the error messages to files named fort.21, fort.22, so it helps to read the message.

2.  Compile the code with all the debugging flags and try the same simulation? 
If you are using intel based compilation:
```
   F_OPTS= -FR -O0 -recursive  -check all,noarg_temp_created -g -debug extended            \
           -debug-parameters all -fpe0 -no-ftz -traceback -ftrapuv -fp-stack-check         \
           -implicitnone -assume byterecl -qopenmp -diag-error=all -static
   C_OPTS= -O0 -DLITTLE  -g -traceback -qopenmp -static
   LOADER_OPTS=-FR -O0 -recursive  -check all,noarg_temp_created -g -debug extended        \
               -debug-parameters all -fpe0 -no-ftz -traceback -ftrapuv -fp-stack-check     \
               -implicitnone -assume byterecl -qopenmp -diag-error=all 
```

or, if you are using gfortran:
```
   F_OPTS= -O0 -ffree-line-length-none -g -fimplicit-none -Wall -finit-real=snan           \
           -finit-integer=-2147483648 -ffpe-trap=invalid,zero,overflow,underflow           \
           -fcheck=all -frecursive -fsignaling-nans -Werror -fopenmp -static
   C_OPTS= -O0 -DLITTLE  -g -static
   LOADER_OPTS= -O0 -ffree-line-length-none -g -fimplicit-none -Wall -finit-real=snan      \
                -finit-integer=-2147483648 -ffpe-trap=invalid,zero,overflow,underflow      \
                -fcheck=all -frecursive -fsignaling-nans -Werror -fopenmp
```

**Comment by femeunier** [comment]
So I changed the location for writing the error message and re-ran the simulations using the debug flags with both intel and gfortran compilers. Here is what I have got:

With intel, it ran fine for the first day just like before, and here is the [kind](https://github.com/EDmodel/ED2/files/4458355/fort.21.txt) of patch budget information that causes the crash.

With gfortran, the crash occurs earlier and here is the [log](https://github.com/EDmodel/ED2/files/4458372/BCI_out.32850489.txt) and the [error](https://github.com/EDmodel/ED2/files/4458374/BCI_err.32850489.txt) files.

**Comment by mpaiao** [comment]
I will think more about this, I don't see any obvious reason.  It does look that there is something missing in the carbon budget, which would indicate that some variable was not properly saved in the history file.  

But just to confirm, are you using a history file produced by the same ED2 version you are trying to resume?

**Comment by femeunier** [comment]
I'll clean everything up and give it another try just to be sure but normally yes ... I'll let you know, thanks for the help anyway!

**Comment by mpaiao** [comment]
@femeunier I just got the exact same error as you in one of my simulations.  I had run the simulation before, and it was working fine if I let the simulation run from beginning to end.  This points to an error loading variables from history the file.

**Comment by femeunier** [comment]
Yes, same here. A run that completed just fine can't be restarted from a previously written restart file. It looks like it was working with the previous pull request so it must be one of these recent changes that cause the crash.

**Comment by mpaiao** [comment]
I don't recall making any change that would affect any of the budget variables, but clearly I did something wrong.  I have to work on something else now, but I will get back to this.

In the meantime, if you have time, try to compare the -S- file you are using to start the run with the -Z- file it creates at the beginning. In principle values should be nearly identical (ideally identical) for every variable. There may be a bunch of false alarms but it may help to spot the variable/variables causing trouble.

**Comment by femeunier** [comment]
I made the difference between those S and Z files and [here](https://github.com/EDmodel/ED2/files/4461636/diff.txt) are the non-zero differences. It looks like there are several significant changes (including EBUDGET_INITIALSTORAGE and CO2BUDGET_INITIALSTORAGE which are the largest).

**Comment by mpaiao** [comment]
@femeunier Yeah, I am trying to figure out what is happening.  

I already found one bug in the history initialization routine

https://github.com/mpaiao/ED2/blob/ddc9f86f18639a0da3f66d6ab8041598739f1f8c/ED/src/io/ed_init_history.f90#L3793

where CBUDGET_RESIDUAL and CBUDGET_COMMITTED were swapped.  I didn't change this recently, so the bug was already there for a while. Unfortunately, fixing this didn't solve the problem for me, so I am still investigating.  Which commit was the last you you successfully ran from history?

**Comment by mpaiao** [comment]
@femeunier I think I fixed the problem.  There were three things happening:
1. HISTORY runs should read as much as possible from the HISTORY file instead of running the initialization routines. This should not cause the problem, but it is better to read the variables from the file.
2. There were two CBUDGET variables that were swapped in ed_init_history.f90.  This was not  causing of the problem in my case, but it would definitely cause problem if anyone had a history file in the middle of the day (rare, but the code should account for the possibility).
3. I had changed the location of where the litter inputs are reset, so I could have these variables in the output.  The problem is that they are also written to the history file and were loaded.  I forced them to be set to zero if the history time is at 00 UTC (daily time step).  My tests seem to be running, when you have a chance, please give it a try too.

**Comment by femeunier** [comment]
Alright, I'll pull the changes and give it a try today.  I'll let you know ASAP

**Comment by femeunier** [comment]
The history restart work just fine now! Thanks for the debugging and the commits. However I still have a few stimulations crashing become wood water potential becomes slightly positive (see  
[log file](https://github.com/EDmodel/ED2/files/4482983/BCI_out.32859353.txt)). Should we bound the maximum values of leaf/wood water potentials (just like the minimum water potentials), which seems to solve the issue or does it indicate a more worrisome problem?

**Comment by xiangtaoxu** [comment]
> The history restart work just fine now! Thanks for the debugging and the commits. However I still have a few stimulations crashing become wood water potential becomes slightly positive (see
> [log file](https://github.com/EDmodel/ED2/files/4482983/BCI_out.32859353.txt)). Should we bound the maximum values of leaf/wood water potentials (just like the minimum water potentials), which seems to solve the issue or does it indicate a more worrisome problem?

This is a little worrisome to me because wood internal water experiences strict boundary check in the rk4 integrator before being copied to patch variables. It is also not the problem of the plant hydraulic calculations otherwise, there will be an error one time step before (projected wood psi is positive)

Could you send me a history file right before the crash and one year's met forcing so that I can reproduce and look into the problem?

**Comment by femeunier** [comment]
Hi Xiangtao! Thanks for jumping in. Here is a [link](https://www.dropbox.com/sh/cliuloja0x2g9ay/AAAPMUdyqkeCImmrgcQttNg_a?dl=0) to reproduce the crash: one year of MET + IC + confix and ED2IN file.The crash occurs the very first timestep of March, 22. Let me know if you have some issue

**Comment by mpaiao** [comment]
There seem to be two issues there, the positive values, and a NaN.  I think the positive value may have some workaround similar to what we did for the minimum value.  The NaN and that it is happening at the beginning of a new day suggests something going wrong with leaf_phenology or growth_balive.  Now that the HISTORY restart is working again, it may be worth restarting the run right before the crash and run with the debugging options.

**Comment by xiangtaoxu** [comment]
I think I find the problem. The NaN value is downstream to the positive wood psi. Will update this response shortly

File: ED/src/dynamics/phenology_driv.f90 [hunk 1]
```diff
@@ -849,7 +849,7 @@ subroutine update_phenology(doy, cpoly, isi, lat,veget_dyn_on)
 
 
                   !----- Adjust plant carbon pools and elongation factor. -----------------!
-                  cpatch%bleaf     (ico) = cpatch%bleaf(ico) - delta_broot
+                  cpatch%broot     (ico) = cpatch%broot(ico) - delta_broot
```
xiangtaoxu, line 852: Good catch!

xiangtaoxu:
I am going to let this sit open for one more day and merge it if no one raises questions. For @femeunier 's problem, I will include the bug fix in my PR scheduled this week

PR #307, opened 2020-06-03
Multiple updates in tropical tree physiology

xiangtaoxu:
This PR contains new plant physiology experimented by Xiangtao Xu for tropical forests, some of which can be useful for other biomes, and some miscellaneous updates. 

**NOTE: This PR needs some more tests to make sure it fully reproduces my personal version.  (To be updated)**

## Description

1. **Rewrite the Katul's optimization stomatal conductance scheme** (dynamics/farq_katul.f90) using Regula false (Illinois algorithm) instead of Newton's methods so that the optimization results are more robust. It is now also more readable and compatible with the existing farq_leuning and accounts for C4 photosynthesis pathway and TPU-limited photosynthesis. Key parameters for different PFTs are updated following _Lin et al. 2015 Nature Climate Change_ and _Manzoni et al. 2011 Functional Ecology_.
2. **Representation of stem maintenance respiration**. Stem respiration in previous versions only come from growth respiration. However, maintenance respiration does happen and might vary with plant size. I add a stand-alone stem respiration module (dynamics/stem_resp_driv.f90) based on _Chambers et al. 2004 Ecological Applications_. Preliminary tests suggest the updates generate better partitioning of autotropic respiration as reported in _Mahli et al. 2008 PTRS B_. This functionality can be turned on and off in ED2IN using `ISTEM_RESPIRATION_SCHEME`.
3. **Updates the light-driven trait plasticity** to allow for cohort-level variations in leaf dark respiration and a new set of parameterization based on data at BCI, Panama. In this scenario (`TRAIT_PLASTICITY_SCHEME` = 3), all PFTs become more shade-tolerant because cohorts can down-regulate Rd under shading.
4. **Include a new `IALLOM` scenario** (4), which is basically the same as IALLOM=3 but assumes leaf area instead of leaf biomass is conserved at a given size when trait plasticity is enabled. Otherwise, small seedlings in deep understory can some times have too much leaf area (SLA more than doubled), leading to a low basal area at the ecosystem level.
5. **Updates tree mortality**. _(a)_ add hydraulic failure mortality based on monthly average xylem PLC, and _(b)_ use the original growth-based mortality reported in Camac et al. 2019 PNAS instead of adapting the relationship using CB as @mpaiao did. Two ED2IN flags are added to ensure compatibility: `CARBON_MORTALITY_SCHEME` (@mpaiao I replace the `ECONOMICS_SCHEME` with this one when it comes to mortality, let me know whether it is OK); `HYDRAULIC_MORTALITY_SCHEME` (highly experimental, needs further tests)
6. **Enable plants to save a target number of carbon in storage for future regular/post-disturbance reflushing instead of using all carbon for structural and reproductive growth**. A new parameter `storage_reflush_times` is added to describe how many times plants can reflush the whole canopy and fine roots by just using storage. This parameter is defined as 2 when phenology is 5 (hydraulics-driven drought deciduous) and 0 otherwise (default values).
7. **Allow for shuffling/cycling meteorology while using a separate long-term CO2 record**, which can be handy for long-term forest spinups. To do this, simply provide CO2 in a separate format that only has CO2 in it. If provided together with other met data, CO2 will be cycled/shuffled.
8. Miscellaneous clean-ups and bug fixes (e.g. hydraulic overcharge reported in #305 ).

TODO: account for new grasses (no stem) in hydraulics [will update in future PRs]

References:
Camac, J. S., R. Condit, R. G. FitzJohn, L. McCalman, D. Steinberg, M. Westoby, S. J. Wright, and D. S. Falster. 2018. Partitioning mortality into growth-dependent and growth-independent hazards across 203 tropical tree species. Proceedings of the National Academy of Sciences of the United States of America 115:12459–12464.

Chambers, J. Q., E. S. Tribuzy, L. C. Toledo, B. F. Crispim, N. Higuchi, J. Dos Santos, A. C. Araújo, B. Kruijt, A. D. Nobre, and S. E. Trumbore. 2004. Respiration from a tropical forest ecosystem: Partitioning of sources and low carbon use efficiency. Ecological Applications 14:S72–S88.
Lin, Y.-S., B. E. Medlyn, R. A. Duursma, I. C. Prentice, H. Wang, S. Baig, D. Eamus, V. R. de Dios, P. Mitchell, D. S. Ellsworth, M. O. de Beeck, G. Wallin, J. Uddling, L. Tarvainen, M.-L. Linderson, L. A. Cernusak, J. B. Nippert, T. W. Ocheltree, D. T. Tissue, N. K. Martin-StPaul, A. Rogers, J. M. Warren, P. De Angelis, K. Hikosaka, Q. Han, Y. Onoda, T. E. Gimeno, C. V. M. Barton, J. Bennie, D. Bonal, A. Bosc, M. Low, C. Macinins-Ng, A. Rey, L. Rowland, S. A. Setterfield, S. Tausz-Posch, J. Zaragoza-Castells, M. S. J. Broadmeadow, J. E. Drake, M. Freeman, O. Ghannoum, L. B. Hutley, J. W. Kelly, K. Kikuzawa, P. Kolari, K. Koyama, J.-M. Limousin, P. Meir, A. C. Lola da Costa, T. N. Mikkelsen, N. Salinas, W. Sun, and L. Wingate. 2015. Optimal stomatal behaviour around the world. Nature Clim. Change 5:459–464.

Manzoni, S., G. Vico, G. Katul, P. A. Fay, W. Polley, S. Palmroth, and A. Porporato. 2011. Optimizing stomatal conductance for maximum carbon gain under water stress: A meta-analysis across plant functional types and climates. Functional Ecology 25:456–467.

Malhi, Y., C. Doughty, and D. Galbraith. 2011. The allocation of ecosystem net primary productivity in tropical forests. Philosophical Transactions of the Royal Society B: Biological Sciences 366:3225–3245.
 
## Collaborators

Marcos Longo @mpaiao 
## Motivation and Context

These updates serve to generate more robust and realistic simulations for tropical forests under plant hydraulics.

## Types of changes

## Checklist:

- [x] I have updated the documentation accordingly.

## Testing :

- [x] All new and existing tests passed.


ED/src/dynamics/farq_katul.f90
```diff
@@ -104,772 +111,676 @@ subroutine katul_lphys(can_prss,can_shv,can_co2,ipft,leaf_par,leaf_temp
...
+         ! scaled with the vm_bar:Vm0 ratio.                                               !
+         !---------------------------------------------------------------------------------!
+         thispft(ib)%vm0  = dble(vm_bar) * umol_2_mol8
+         thispft(ib)%rd0  = dble(rd_bar) * umol_2_mol8
+         f_plastic8       = dble(vm_bar) / dble(vm0(ipft))
+         thispft(ib)%jm0  = f_plastic8 * dble(jm0 (ipft)) * umol_2_mol8
+         thispft(ib)%TPm0 = f_plastic8 * dble(TPm0(ipft)) * umol_2_mol8
+         !---------------------------------------------------------------------------------!
+      case default
+         !------ No plasticity: use default parameters. -----------------------------------!
```
mpaiao, line 248: @xiangtaoxu I was checking this part of the code (and the equivalent one in farq_leuning.f90), and I think we should eliminate this select case block here and in farq_leuning.f90, and use only the "within-canopy" trait plasticity case.  In the case of light-controlled leaf phenology, vm_bar is not vm0, it is set by the site-level variable cpoly%vm_bar_toc(ico).  I thought I had changed this, but apparently I forgot to do it.

If both trait plasticity and light-controlled phenology are disabled, then vm_bar is equal to vm0, so it would be equivalent to the "case default".  If you and others are fine with this, I can submit a PR right after merging this.

xiangtaoxu: I think you are right. I have updated this pr accordingly for both farq_katul.f90 and farq_leuning.f90

ED/src/dynamics/structural_growth.f90
```diff
@@ -1121,7 +1152,17 @@ subroutine plant_structural_allocation(ipft,hite,dbh,lat,phen_status,elongf,bdea
...
-      f_bstorage = 1.0 - f_growth - f_bseeds
+
+      ! plants will save bstorage_reserve as mobile carbon supply for potential reflushing
+      ! we take the maximum of the residual of f_bseeds and f_growth and bstorage_reserve 
+      ! / bstorage to maintain compatibility with previous versions (bstorage_reserve = 0.)
+      f_bstorage = max(1.0 - f_bseeds - f_growth, bstorage_reserve / bstorage)
+
+      ! we need to modify f_bseeds and f_growth accordingly if f_bseeds+f_growth is non-zero
+      if (f_bseeds + f_growth > 0.0) then
+          f_bseeds = f_bseeds / (f_bseeds + f_growth) * (1.0 - f_bstorage)
```
mpaiao, line 1163: @xiangtaoxu I think we should do the following instead:
```
f_bgi    = 1.0 / (f_bseeds + f_growth)
f_bseeds = f_bseeds * f_bgi * (1.0 - f_bstorage)
f_growth = f_growth * f_bgi * (1.0 - f_bstorage)
```

Otherwise f_bseeds is already modified by the time you update f_growth and the sum may not be 1.0.

ED/src/init/ed_params.f90
```diff
@@ -3536,9 +3528,19 @@ subroutine init_pft_alloc_params()
             !    Amazon.  Global Biogeochem. Cycles, 30(11):1639-1660.                     !
             !    doi:10.1002/2016GB005465 (L16).                                           !
             !------------------------------------------------------------------------------!
-            b1Bl(ipft) = c14f15_bl_xx(1) / SLA(ipft)
+            b1Bl(ipft) = c14f15_bl_xx(1) / SLA(ipft) ! XX --> MLO: should ther be a C2B here given c14f15_bl_xx is in m2 (?) and SLA is m2/kgC
```
mpaiao, line 3531: @xiangtaoxu I need to double check this, but I think I took this into account. The C2B is applied in allometry_utils.f90, to be consistent with the previous allometric equations.

ED/src/init/ed_type_init.f90
```diff
@@ -168,9 +169,16 @@ subroutine init_ed_cohort_vars(cpatch,ico, lsl,mzg,ntext_soil)
          cpatch%sla   (ico) = SLA(ipft)
          !---------------------------------------------------------------------------------!
       end select
+      ! Always use tabulated values for rd_bar
+      ! XX->MLO Check whether this is OK for your trait economics scenario
```
mpaiao, line 173: @xiangtaoxu I think this is consistent with what was done before. Trait economics shouldn't matter here, the select case block above is for the light-controlled phenology.

An alternative would be to set the following:
```
catch%rd_bar(ico) = Rd0(ipft) * catch%vm_bar(ico) / Vm0 (ipft)
```
but I don't think this would be consistent with the light phenology paper.

mpaiao:
@xiangtaoxu I checked your changes and I think almost everything looks good.  There is just the one remark in structural growth that I think it's a minor bug.

ED/src/dynamics/structural_growth.f90
xiangtaoxu: Good catch! Thanks!

xiangtaoxu:
@mpaiao thanks for the review. I will update the pr soon!

xiangtaoxu:
@mpaiao I have updated the bug fix. Feel free to merge this pr so that Erik can start work on merging his version. I will include some minor updates and test suites later

PR #331, opened 2021-06-28
[Bug fix] Fix errors about low leaf water potential in plant hydraulics

xiangtaoxu:
Simulations with plant hydraulics can lead to infrequent low leaf water potential errors and model crash for drier sites when drought-phenology is turned on. This is because (1) the hydraulic module compares wood water potential with leaf_psi_min when leaves are off and (2) cohorts will not be able to absorb water when all fine roots are lost during long dry seasons even when wet season starts again. 

## Description

1. Update plant_hydro.f90 to use correct metric when determining whether water potential is too low
2. Add a new parameter eSRAc represents the equivalent specific root area from coarse roots. It is calculated assuming coarse roots are ~10cm in diameter on average. In total the added constant water conductance is 1-5% of water conductance from fine roots. Therefore, the new addition will not change maximum soil-root water conductance much but will enable an initial water flux into root when wet season starts even if all the fines are shed due to drought-phenology or carbon starvation.

All changes should be contained within the plant hydro module and should be mostly orthogonal to other on-going changes.

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue) 

## Checklist:

## Testing :

### DIFF
--- a/ED/src/dynamics/plant_hydro.f90
+++ b/ED/src/dynamics/plant_hydro.f90
@@ -58,7 +58,11 @@ module plant_hydro
       use physiology_coms      , only : plant_hydro_scheme     ! ! intent(in)
       use pft_coms             , only : C2B                    & ! intent(in)
                                       , leaf_water_cap         & ! intent(in)
+                                      , wood_water_cap         & ! intent(in)
+                                      , SRA                    & ! intent(in)
+                                      , eSRA_c                 & ! intent(in)
                                       , leaf_psi_min           & ! intent(in)
+                                      , wood_psi_min           & ! intent(in)
                                       , small_psi_min          ! ! intent(in)
 
       implicit none
@@ -82,6 +86,8 @@ module plant_hydro
       real                                :: bsap        !< sapwood biomass        [    kgC]
       real                                :: transp      !< transpiration rate     [   kg/s]
       real                                :: c_leaf      !< leaf capacitance       [   kg/m]
+      real                                :: c_stem      !< stem capacitance       [   kg/m]
+      real                                :: RAI_total   !< total root area index  [   m2/m2]
       logical                             :: track_hydraulics !< whether track hydraulics
       !----- Variables for debugging purposes ---------------------------------------------!
       integer, parameter                  :: dco        = 0 ! the cohort to debug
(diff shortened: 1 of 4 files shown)


mdietze:
I'm OK with the concept, but am wondering about the calibration. First, given that a coarse root is anything >2mm, the default of 10cm seems really big -- almost all coarse roots are way smaller than this and many tree size classes won't have any roots this size at all. Furthermore, I doubt roots this size are anything other than structural. Second, where does the 0.01 multiplier come from for the uptake? Is this completely ad hoc or based on something in the literature or the root databases (e.g. FRED).

xiangtaoxu:
> I'm OK with the concept, but am wondering about the calibration. First, given that a coarse root is anything >2mm, the default of 10cm seems really big -- almost all coarse roots are way smaller than this and many tree size classes won't have any roots this size at all. Furthermore, I doubt roots this size are anything other than structural. Second, where does the 0.01 multiplier come from for the uptake? Is this completely ad hoc or based on something in the literature or the root databases (e.g. FRED).

Yes, the current approach is largely ad hoc. The purpose is to allow plans to 'sense' soil wetting even if they don't have active fine roots. I use coarse root surface area to give me a baseline value that can be calculated for each cohort and each soil layer. The 0.01 multiplier is a tuning parameter to indicate coarse roots are not as permeable as fine roots so that water conductance from fine roots are still dominant in soil-water water flux.

I agree linking it to measurable root traits would be great (I will check FRED later). Currently, I treat this as a fix for a rarely occurring chicken-and-egg bug (e.g. trees with enough bstorage lose all their fine roots during a long dry season. When rainy season comes, they need fine root to absorb water and increase water potential but phenology scheme says they can only grow fine roots when water potential is high enough).

mpaiao:
@xiangtaoxu I agree with @mdietze on adding more parameters that we don't have a good way to constrain. Also, I wonder about assuming that coarser roots have an average diameter of 10 cm. Shouldn't this be scaled with the cohort size?

mpaiao:
Also, notice that the CI test for Tonzi with harvesting failed because of a floating point exception in events.f90.  This may be indicating that some variable has not been properly initialised, you may want to check that too. Thanks!

mpaiao:
@xiangtaoxu Just checking in about this pull request, is there any update on it? Should we merge it?

xiangtaoxu:
No need to merge esp. because it is not consistent with the current version. My lab will create a new PR later this year bundling all ecophysiology updates

mpaiao:
Sorry for not following up sooner... I will close this PR then.

PR #341, opened 2022-02-21
Reduce maximum dimensions for Mac OS

mpaiao:
The maximum dimensions of several properties exceed the maximum stack memory usage in Mac OS X. Assuming that most use of ED2 with Mac OS X is for small simulations, I added preprocessor instructions to ed_max_dims.F90 to make these dimensions substantially lower when the machine is set to MAC_OS_X. With these minor edits, I was able to successfully compile and run ED2 on my Mac.

These edits should not have any impact on other systems, as long as people are not setting CMACH=MAC_OS_X to compile ED2 on non-Mac systems (which they shouldn't be doing in any case).

## Types of changes

## Checklist:

## Testing :


mpaiao:
It seems the continuous integration has been failing for a while now. I checked the log and it looks like the failure is occurring during the apt-get stage. I haven't looked into this for a while, does anyone have suggestions on how to fix it? Thanks!

PR #350, opened 2022-04-28
Skip derivation of light phenology parameters

mpaiao:
When IPHEN_SCHEME≠3, radiation controls on phenology are disabled, and the code should bypass all calculations. 
The code would still calculate derived parameters, which could lead to floating point exceptions depending on the ED2IN/xml settings. This fix sets the minimum and maximum radiation to the solar constant in ed_params.f90. This will also avoid turnover amplitude to be calculated in in sub-routine update_turnover when light phenology is disabled.

## Types of changes

## Checklist:

## Testing :


mpaiao:
Merging the pull request because this is a minor bug fix.

PR #372, opened 2022-11-14
Update radiate_driver.f90

Lo-hatch:
a mistake when weighting surface water albedo by water fraction.  In the old equation, when water fraction is 1, the albedo is **snow albedo**, which should have been **damp soil albedo**

## Types of changes

## Checklist:

## Testing :

### DIFF
--- a/ED/src/dynamics/radiate_driver.f90
+++ b/ED/src/dynamics/radiate_driver.f90
@@ -714,10 +714,14 @@ module radiate_driver
          !      Community Land Model (CLM). NCAR Technical Note NCAR/TN-478+STR.           !
          !                                                                                 !
          !---------------------------------------------------------------------------------!
-         albedo_sfcw_par = albedo_damp_par + csite%sfcwater_fracliq(ksn,ipa)               &
-                                           * ( snow_albedo_vis - albedo_damp_par )
-         albedo_sfcw_nir = albedo_damp_nir + csite%sfcwater_fracliq(ksn,ipa)               &
-                                           * ( snow_albedo_nir - albedo_damp_nir )
+         !albedo_sfcw_par = albedo_damp_par + csite%sfcwater_fracliq(ksn,ipa)               &!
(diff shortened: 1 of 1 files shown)


mdietze (approved):
Agree that the math was backwards. 

That said, I'm not sure the model actually makes sense to me. Melting snow is going to continue to have high reflectivity until it's almost gone -- I think the logic here makes more sense if you're thinking about the fraction AREA of snow than the fractional liquid state of a melting snow. Likewise, in the rare occasions that there is persistant fully-liquid surface water (i.e. ponding, flooding, wetlands), wouldn't we want to use the albedo of liquid water not wet soil?

Lo-hatch:
> Agree that the math was backwards.
> 
> That said, I'm not sure the model actually makes sense to me. Melting snow is going to continue to have high reflectivity until it's almost gone -- I think the logic here makes more sense if you're thinking about the fraction AREA of snow than the fractional liquid state of a melting snow. Likewise, in the rare occasions that there is persistant fully-liquid surface water (i.e. ponding, flooding, wetlands), wouldn't we want to use the albedo of liquid water not wet soil?

Yes, I agree with you about the snow faction, but the model indeed use "snowfac" just after the code I modified (line736-737) to calculate the absorbed radiation.
     sfcw_odepth   = min( lnexp_max,   20.0 * csite%sfcwater_depth(k,ipa)   / csite%snowfac   (ipa) )
however, I know little about the physical process of radiation and not sure if it is ok to use "snowfac" like that.
I also agree with you about the permanent surface water albedo, but It seems that ED only allows temporary surface water? Correct me if I am wrong. Thanks!

mpaiao:
Thanks for spotting this bug, @Lo-hatch! 
@mdietze I agree with you although I have the impression that the ponding albedo would depend on the depth of the surface water. To be honest the entire radiative transfer of snow (and temporary surface water) in ED2 could use some upgrade at some point.

PR #390, opened 2023-06-12
Minor edits to re-enable CI testing (plus post-processing updates)

mpaiao:
Disabled MPI from CI due to likely issue with MPI itself (not with ED2), and some minor code fixes so CI successfully runs again.

## Description

1. Replaced ```include 'mpif.h'``` statements with ```use mpi``` so the code is aligned with Fortran usage.
2. Removed MPI settings from include.mk.travisci. This may be temporary, but the code was crashing when some of the MPI functions were used multiple times in the same ED2 sub-routine, but with different elements (e.g., scalar, vector, array, or integers and real numbers).  The ED2 code is using the functions as they should be used, so it seems the problem is on the MPI side. This solution may be temporary. A full solution might require adding wrappers to the MPI calls (or some fixes in the MPI library source code).
3. Updated a bunch of post-processing scripts, which should not affect the code itself. 

## Types of changes

## Checklist:

## Testing :


mpaiao:
@robkooper @pmoorcroft 
It seems the Docker build is still failing, and I can imagine it's due to similar issues... The interface issue seems to be on the MPI side (as opposed to a bug in ED2), so I don't know what the best approach is, other than disabling MPI for now...

PR #411, opened 2025-01-02
ED2IN clean up

mpaiao:
Deleted several variables that used to modify PFT and turbulence parameters from ED2IN.

## Description

Several variables in ED2IN modified parameters (mostly PFT parameters), which conflicted with the XML approach of modifying them. The code was streamlined so parameters are now exclusively managed by the XML interface. 

Another related change is that `NL%IPHEN_SCHEME` no longer defines the leaf phenology strategy: these now must be set through XML (parameter `phenology`).  Instead, `NL%IPHEN_SCHEME` simply defines whether to use the predicted cold-deciduous phenology (`NL%IPHEN_SCHEME = 0`) based on [Botta et al. (2000)](https://dx.doi.org/10.1046/j.1365-2486.2000.00362.x), or the prescribed leaf phenology (`NL%IPHEN_SCHEME = 1`) described in [Medvigy et al. (2009)](https://dx.doi.org/10.1029/2008JG000812).

## Motivation and Context

This change is long due and reduces confusion between the roles of ED2IN (model approaches and site-specific quantities) and the xml files (PFT-specific or site-independent parameters). 

Some turbulence-related parameters (not PFT-specific) required a different approach, because they are defined through namelist when running in coupled mode. In this case, XML still has the last word in ED2, but if they are not set from XML, then the code prioritise the values from RAMSIN over the ED2 defaults.

As a minor change, some non-standard characters in comments were removed from the code, to improve portability to text editors.

## Types of changes

## Checklist:

Changes to documentation will be applied once the pull request is approved and merged.

## Testing :


mpaiao:
@mdietze @xiangtaoxu when you have a chance, would one of you mind checking this PR and a few others I submitted in Dec/Jan. I can then revise a few others to try to clear the PR queue. Thanks!

xiangtaoxu:
Hi, Marcos. I can come to ED2 code review likely next week after end-of-semester stuff.... Thanks for cleaning-up the codes!

PR #414, opened 2025-02-11
Minor code changes to allow CI tests to run

mpaiao:
## Description

Not sure why this was not working, but in `ed_read_ed10_20_history.f90` the following construction was leading to an FPE signal:

```fortran
   leaf_lifespan(:) = merge( 12.0 / leaf_turnover_rate(:)                                  &
                           , llspan_inf                                                    &
                           , leaf_turnover_rate(:) > 0.0  )
```

I replaced it with the equivalent `where` construction

```fortran
   where (leaf_turnover_rate(:) > 0.0)
      leaf_lifespan(:) = 12.0 / leaf_turnover_rate(:)
   elsewhere
      leaf_lifespan(:) = llspan_inf
   end where
```

This change seems to have fixed the problem. This could be a compiler issue, but if the problem pops up again, we may want to deprecate `merge` in the code and replace it with `where` instead.

I also updated the test suite script `run-test.sh` to pull data from the EDModel organization instead of downloading from a specific user. The datasets are identical to the ones @ashiklom kindly hosted over the years, but having this within the EDModel seems more logical in case we want to add more tests or change them in the future. Hope this is fine with you @ashiklom!

## Collaborators

@ashiklom

## Motivation and Context

This fixes crashes in the CI.

## Types of changes

## Checklist:

## Testing :

### DIFF
--- a/ED/src/io/ed_read_ed10_20_history.f90
+++ b/ED/src/io/ed_read_ed10_20_history.f90
@@ -179,9 +179,11 @@ subroutine read_ed10_ed20_history_file
    !---------------------------------------------------------------------------------------!
    !     Define PFT-dependent leaf life span, used for initialisation.                     !
    !---------------------------------------------------------------------------------------!
-   leaf_lifespan(:) = merge( 12.0 / leaf_turnover_rate(:)                                  &
-                           , llspan_inf                                                    &
-                           , leaf_turnover_rate(:) > 0.0  )
+   where (leaf_turnover_rate(:) > 0.0)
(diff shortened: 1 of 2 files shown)


ashiklom (approved):
Looks good to me! It's been forever, so I forgot how all of this works...so I'm quite happy to not be directly responsible for this anymore ;)

mpaiao:
Thanks Alexey! I don't remember very well either ;).

By the way, these tests will crash because the meteorological drivers in EDModel are already expecting pull request #413... Apologies for the mix-up!

PR #415, opened 2025-02-21
added ISLCOLFLG to output

sdeherto:
Made small code changes in ED/src/memory/ed_state_vars.F90 to add the variable ISLCOLFLG to output as this was causing downstream error when applying postprocessing script in R-Utils


mpaiao (approved):
Thanks for the pull request, @sdeherto. Your changes look good with me!

Issue #436, opened 2026-06-11
Fortran runtime error: Index '143' of dimension 1 of array 'cpu_use' above upper bound of 64


LMurphy186232:
I recently compiled ED2 for use on a new-to-me HPC system. I got the following error at runtime:

At line 200 of file edmain.F90 Fortran runtime error: Index '143' of dimension 1 of array 'cpu_use' above upper bound of 64
with an unhelpful backtrace. I investigated this error, and I think I know what is happening.

This is the code in question in edmain.F90, which starts at line 178:

   !---------------------------------------------------------------------------------------!
   !     Check OMP thread and processor use and availability.  This is done outside the    !
   !  preprocessor "if block" because it is possible to activate multithread without       !
   ! compiling with mpif90.                                                                       !
   !                                                                                       !
   ! Note: One could use omp_get_num_threads() in loop, but that would depend on how many  !
   ! threads were open at the time of its call.                                            !
   !---------------------------------------------------------------------------------------!
   max_threads   = 1
   num_procs     = 1
   thread        = 1
   cpu           = 1
   thread_use(:) = 0
   cpu_use(:)    = 0
   !$ max_threads = omp_get_max_threads()
   !$ num_procs   = omp_get_num_procs()

   !$OMP PARALLEL DO DEFAULT(SHARED) PRIVATE(thread,cpu)
   do n = 1,max_threads
     !$ thread = omp_get_thread_num() + 1
     !$ cpu    = findmycpu() + 1
     thread_use(thread) = 1
     cpu_use(cpu)       = 1
   end do
   !$OMP END PARALLEL DO
   !---------------------------------------------------------------------------------------!

 The line "cpu    = findmycpu() + 1" appears to be returning a CPU ID number, which may be larger than the "cpu_use" max array length of 64 on an HPC system. The array overrun is causing an error to be thrown.

I do not have the OMP coding chops to write a pull request to fix this, unfortunately, so I am filing a bug report instead. This appears to be needed for an informational banner (only?), so I just commented it out in my version, and I am back up and running.

xiangtaoxu:
Thanks, Lora! Have requested more than 64 cores for your job? Are you able to run without error if you request <64 cores in your HPC system?

mpaiao:
Thanks for flagging this @LMurphy186232. If I recall correctly, this OMP code was written by @rgknox a long time ago. Looking back at it, I wonder if the solution should be simply to turn both thread_use and cpu_use into allocatable arrays.

[...]

PR #439, opened 2026-07-30
Improving counting of threads and CPUs for MPI jobs

LMurphy186232:
Fix for array overruns when reporting MPI core usage.

## Description

Changed the reporting of cores and threads in MPI runs to count core and thread IDs in dynamically allocated arrays, allowing for correct reporting no matter how these IDs are assigned.

Previously, the code assumed that when reporting on core and thread assignments to ED2 in MPI runs, ID numbers would be numbered sequentially starting at 0, and that there would be a maximum of 64 of each. Current HPC systems break both of those assumptions and this led to occasional array overflows.

## Collaborators

@xiangtaoxu 
@mpaiao 

## Types of changes

- [ ] Hot fix (emergential bug fix to make ED2 run again)
- [x] Bug fix (non-breaking change which fixes an issue) [https://github.com/EDmodel/ED2/issues/436] 

### Changes in Settings, Input Files or Output Files
  - [ ] This feature requires changes in ED2IN
  - [ ] This feature requires changes in the XML parameter file
  - [ ] This feature requires changes in other input files
  - [ ] This feature will change output files

 

### Expectation of Answer Changes:

- [ ] No changes expected (bit-for-bit compatibility)
- [X] Changes expected with specific configurations (see details below)
- [ ] Changes expected in all simulations (see details below)

No change expected in model output, just in the capacity of certain runs to proceed at all.

## Checklist:

- [ED2 Wiki](https://github.com/EDmodel/ED2/wiki) update:

## Testing :

### DIFF
--- a/ED/src/driver/edmain.F90
+++ b/ED/src/driver/edmain.F90
@@ -38,6 +38,9 @@ program main
    integer                               :: nslaves
    integer                               :: isingle
    integer                               :: n
+   integer                               :: m
+   integer                               :: used_threads
+   integer                               :: used_cpus
    !------ Local variables (MPI only). ----------------------------------------------------!
 #if defined(RAMS_MPI)
    integer                               :: ierr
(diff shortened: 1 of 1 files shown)


mpaiao (approved):

mpaiao:
Thanks for PR @LMurphy186232, I went through it and it looks good to me.

yixin98:
I tested this on a cluster node with 96 cores, and it resolved the array overflow error that used to occur.

